PR #17, opened 2020-09-24
Sanitize CLI Input

ryandgoulding:
There are several clients that form CLI command strings using fmt.Sprintf(...).
Although this utility is great, it lacks in sanitization and input checking.
The raw input is trusted and spewed to the output string.

This change introduces a "utils" package which contains common utilities for
consumption from both generic and cnf-specific test cases.  The first utility
method provided is "PrepareString".  PrepareString is a lightweight wrapper
around fmt.Sprintf(...) which checks that input arguments post-trim do not
contain whitespace.  The API signature fmt.Sprintf(string, ...interface{})
were not appropriate here, since we are just substituting strings;  thus
PrepareString(string, ...string) is used instead.


ryandgoulding:
It would be easier to rewrite the code at this point than try to rebase.  I will close this out for now.

PR #98, opened 2021-03-02
Add flag to control claim output location

crwr45:

### DIFF
--- a/test-network-function/suite_test.go
+++ b/test-network-function/suite_test.go
@@ -41,8 +41,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


test-network-function/suite_test.go
```diff
@@ -41,8 +41,11 @@ import (
 )
 
 const (
+	claimFileName                 = "claim.json"
```
ryandgoulding, line 44: Nothing to do for this PR, but in the future we should probably generate this instead with some sort of timestamp to prevent overwriting past attempts.

ryandgoulding (approved):

PR #100, opened 2021-03-02
Containerised tests

crwr45:
Adds a dockerfile, helper, and docs for running the tests in a container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ RUN make install-tools && \
 	make update-deps && \
 	make build-cnf-tests
 
-#  Extract what's needed to run to a seperate location
+#  Extract what's needed to run at a seperate location
 RUN mkdir ${TNF_BIN_DIR} && \
 	cp run-cnf-suites.sh ${TNF_DIR} && \
 	cp version.json ${TNF_DIR} && \
@@ -75,4 +75,4 @@ FROM scratch
 COPY --from=build / /
 ENV KUBECONFIG=/usr/tnf/kubeconfig/config
 WORKDIR /usr/tnf
-CMD ["./run-cnf-suites.sh", "-o", "claim", "generic"]
+CMD ["./run-cnf-suites.sh", "-o", "claim", "diagnostic", "generic"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,78 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest AS build
+
+# golang 1.14.12 as it's the only version available on yum at time of writing.
+ENV GOLANG_VERSION=1.14.12
+ENV GOLANGCI_VERSION=v1.32.2
+
+ENV TNF_DIR=/usr/tnf
+ENV TNF_SRC_DIR=${TNF_DIR}/tnf-src
+ENV TNF_BIN_DIR=${TNF_DIR}/test-network-function
+
+ENV TEMP_DIR=/tmp
+# Most recent version of 4.4 available at time of writing.
+ENV OC_SRC_ARCHIVE=openshift-clients-4.4.0-202006211643.p0.tar.gz
+ENV OC_SRC_URL=https://github.com/openshift/oc/archive/${OC_SRC_ARCHIVE}
+ENV OC_SRC_DIR=${TEMP_DIR}/oc-client-src
+
+# Install dependencies
+RUN yum install -y golang-${GOLANG_VERSION} jq make git
+
+# Build oc from source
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+
+# golangci-lint
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /usr/bin ${GOLANGCI_VERSION}
+
+# Add go binary directory to $PATH
+ENV PATH="/root/go/bin:${PATH}"
+
+# This step is identical to `make install-tools` from the Makefile, but there
+# are version issues if the command is run from within the `/tnf` directory.
+# TODO: resolve the version issues to make it work with line ~50
```
ryandgoulding, line 36: What do you mean?  Can you please share the error you receive?

This is indicative of version as `go get` cannot pin a particular release tag.  It would indicate that either:
1) We are pulling in an older version elsewhere.
2) We should vendor the tools instead of dynamically grabbing the most recent release tag.

Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+RUN git clone --depth=1 --branch=${TNF_VERSION} https://github.com/redhat-nfvpe/test-network-function ${TNF_SRC_DIR}
+
+# Build TNF binary
+WORKDIR ${TNF_SRC_DIR}
+# TODO: RUN make install-tools
+RUN make mocks && \
+	make update-deps && \
+	make build-cnf-tests
+
+#  Extract what's needed to run to a seperate location
```
ryandgoulding, line 54: "to run at"

Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+	rm -rf /usr/lib/golang/pkg && \
+	rm -rf /usr/lib/golang/src
+
+# Copy the state into a new flattened image to reduce size.
+# TODO run as non-root
+FROM scratch
+COPY --from=build / /
+ENV KUBECONFIG=/usr/tnf/kubeconfig/config
+WORKDIR /usr/tnf
+CMD ["./run-cnf-suites.sh", "-o", "claim", "generic"]
```
ryandgoulding, line 85: Does this pin us only to the `generic` test spec?  Shouldn't we take CLI arguments here instead?

ryandgoulding (changes requested):

ryandgoulding (approved):

run-container.sh
```diff
@@ -0,0 +1,62 @@
...
+if (($# < 7)); then
+	usage_error
+fi;
+
+# Parge args beginning with -
+while [[ $1 == -* ]]; do
+	echo $1 $2
+    case "$1" in
+      -h|--help|-\?) usage; exit 0;;
+      -k) if (($# > 1)); then
```
ryandgoulding, line 38: As a comment but not a requirement, we could make `-k` optional.  My idea is that if `-k` is specified, then it takes precedence.  If `-k` is not specified, `$KUBECONFIG` is fed in from the underlying host.  If `-k` is not specified, and `$KUBECONFIG` is not defined, then it is an error case.  This may be a story we could create and add to the backlog for later if time permits.

README.md
```diff
@@ -8,6 +8,30 @@ Container Platform.
...
+
+To run the tests in the container, use the following command. There are several required arguments:
+
+* `-k` gives a path to the kube config file to be used by the container to authenticate with the cluster.
+* `-t` gives the local directory that contains tnf config files set up for the test.
+* `-o` gives the local directory that the test results will be available in once the container exits.
+* Finally, the specs to be run must be specified.
+
+```shell-script
+./run-container.sh -k ~/.kube/config -t ~/tnf/config -o ~/tnf/output diagnostic generic
```
ryandgoulding, line 23: We may want to consider making diagnostic enabled by default, and adding a flag to disable.  That said, that is work for a future version.

Dockerfile
crwr45: This only comes into effect if the container is run without specifying the command. I wanted to have something here, but it's not the intended use case. The docs and run script specify their own run command.
I'll add "diagnostic" as well, but the CMD here serves largely as documentation of one way to run the tests.

run-container.sh
crwr45: Absolutely. I considered the same but decided it was an unnecessary step at this point given time spent on the story. I'll document it in a new story.

Dockerfile
crwr45: addressed in #105

ryandgoulding (approved):

ryandgoulding (approved):

PR #111, opened 2021-03-08
add Turnium test for link status

crwr45:
Also add license to `diagnostic` package

### DIFF
--- a/test-network-function/turnium/suite.go
+++ b/test-network-function/turnium/suite.go
@@ -14,15 +14,15 @@
 // with this program; if not, write to the Free Software Foundation, Inc.,
 // 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-package diagnostic
+package turnium
 
 import (
-	"encoding/json"
 	"path"
 	"time"
 
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/gomega"
+	tnfConfig "github.com/redhat-nfvpe/test-network-function/pkg/config"
 	"github.com/redhat-nfvpe/test-network-function/pkg/tnf"
 	"github.com/redhat-nfvpe/test-network-function/pkg/tnf/handlers/generic"
 	"github.com/redhat-nfvpe/test-network-function/pkg/tnf/interactive"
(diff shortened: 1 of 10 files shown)


pkg/config/generic.go
```diff
@@ -171,6 +171,7 @@ type Container struct {
 
 // TestConfiguration provides generic test related configuration
 type TestConfiguration struct {
+	TurniumBonder       ContainerIdentifier               `yaml:"turniumBonder" json:"turniumBonder"`
```
ryandgoulding, line 174: This should be in its own Turnium-specific configuration `struct` and registered with the `ConfigPool`. `generic` config is for the generic test suite.

ryandgoulding: Additionally, exported struct elements should be documented (even though they aren't for others below)...

test-network-function/turnium/turnium.json
```diff
@@ -0,0 +1,23 @@
+{
+	"identifier": {
+	  "url": "http://test-network-function.com/tests/turnium",
```
ryandgoulding, line 3: This needs to be documented manually in `identifiers.go`.

test-network-function/turnium/turnium.json
```diff
@@ -0,0 +1,23 @@
...
+	  "version": "v1.0.0"
+	},
+	"description": "Check `legids -v` reports state as up.",
+	"testResult": 0,
+	"testTimeout": 10000000000,
+	"reelFirstStep": {
+	  "execute": "legids -v\n",
+	  "expect": [
+		"state: up"
```
ryandgoulding, line 12: We should probably add logic for `state: down`.  After all, state down implies a fail rather than an error (which is the default test result).

test-network-function/turnium/suite.go
```diff
@@ -0,0 +1,97 @@
...
+	// relativeSchemaPath is the relative path to the generic-test.schema.json JSON schema.
+	relativeSchemaPath = path.Join(pathRelativeToRoot, schemaPath)
+
+	// schemaPath is the path to the generic-test.schema.json JSON schema relative to the project root.
+	schemaPath = path.Join("schemas", "generic-test.schema.json")
+)
+
+// createOc sets up a local shell expect.Expecter, checking errors along the way.
+func createOc(pod, container, namespace string) *interactive.Oc {
+	oc, _, err := interactive.SpawnOc(interactive.CreateGoExpectSpawner(), pod, container, namespace, defaultTestTimeout, expect.Verbose(true))
```
ryandgoulding, line 61: Should be fine as I believe the suite is run in a separate thread.  However, we will want to test this with all other test specs.

ryandgoulding (changes requested):

PR #112, opened 2021-03-09
Make jsontest-cli arg ordering match the underlying method

crwr45:
the `jsontest-cli oc` tool is intended for running single json-defined tests, to aid debugging and testing. However, it requires the `namespace`, `pod` and `container` arguments in a different order to the goland method used to spawn an OC connection. This is potentially confusing and frustrating.

This change makes the smaller change to resolve this: modifying the command line args to match the underlying method.
Alternatively, the other approach could be taken but would require a more wide-reaching set of changes.


crwr45:
Fix this the other way round; change the method not the CLI

PR #113, opened 2021-03-09
refresh test result constants

crwr45:
The constants that define test results were defined as
`iota` values, in alphabetical order. This means that a
result of `0` was considered to be an error. The ordering is
very likely the result of a linting rule being applied
overzelously.

The use of both `tnf.ERROR` and other results as the value
for calls to `os.Exit` strongly implies that the `SUCCESS`,
i.e. test pass, should be `0`, indicating a successful execution
and allowing for sensible use in CI pipelines.

This change addresses the issue by individually defining the
constants and updating documentation, comments, unit tests, and
JSON-defined CNF tests to use the new values.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -120,9 +120,9 @@ A human-readable description is required for every test.  This documentation pro
 
 #### testResult
 
-`testResult` is initialized as 0.  As of now, `tnf.Tester` mandates `Result()` returns an `int` value.  `0` corresponds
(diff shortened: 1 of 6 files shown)


crwr45:
this was done on purpose.  So the deal is... we should start tests in tnf.ERROR state until they can prove they have worked.  With the JSON test framework, we do MarshalJSON() the JSON input.  if testResult is not specified in the JSON payload, it will initialize to 0.

The `os.Exit` calls are the actual error, fix it by mapping the `testResult` to an `exitCode`.

PR #114, opened 2021-03-10
add a mapping for test result code to Unix return code

crwr45:
A recent change to use `0` as the ERROR value for test result meant
that some components were reporting a success when they had failed.

This change adds a map to transform the value of test result to a
meaningful Unix exit code.

### DIFF
--- a/cmd/oc/main.go
+++ b/cmd/oc/main.go
@@ -29,16 +29,24 @@ import (
 	expect "github.com/ryandgoulding/goexpect"
 )
 
+const (
+	// incorrectUsageExitCode is the Unix return code used when the supplied arguments are inappropriate.
+	incorrectUsageExitCode = 2
+
+	// mandatoryNumArgs is the number of positional arguments required.
+	mandatoryNumArgs = 3
+)
+
 func parseArgs() (*interactive.Oc, <-chan error, string, time.Duration, error) { //nolint:gocritic //permit unnamed return values
 	timeout := flag.Int("t", 2, "Timeout in seconds")
 	flag.Usage = func() {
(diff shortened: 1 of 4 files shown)


cmd/oc/main.go
```diff
@@ -29,16 +29,24 @@ import (
 	expect "github.com/ryandgoulding/goexpect"
 )
 
+const (
+	// incorrectUsageExitCode is the Unix return code used when the supplied arguments are inappropriate.
+	incorrectUsageExitCode = 2
+
+	// mandatoryNumArgs is the number of positional arguments required
```
ryandgoulding, line 36: Please end with a period to be consistent with the comment for `incorrectUsageExitCode`.

pkg/tnf/test.go
```diff
@@ -33,6 +33,13 @@ const (
 	FAILURE
 )
 
+// EXITCODEMAP maps a test result value to a more appropriate Unix return code.
+var EXITCODEMAP = map[int]int{
```
ryandgoulding, line 37: Please use the convention of MixedCaps as described here: https://golang.org/doc/effective_go#mixed-caps.

I.e.,

`ExitCodeMap`

pkg/tnf/test.go
```diff
@@ -33,6 +33,13 @@ const (
...
+	SUCCESS: 0,
```
ryandgoulding, line 38: For constants, the convention is not clear in Effective Go:  https://golang.org/doc/effective_go#constants.  Thus, let's just keep them as is for now.

ryandgoulding (approved):

PR #126, opened 2021-03-22
Test CNF pod namespace

ydayagi:
Must not begin with 'openshift-' and can not be equal to 'default'

https://issues.redhat.com/browse/CTONET-780

### DIFF
--- a/test-network-function/generic/suite.go
+++ b/test-network-function/generic/suite.go
@@ -156,6 +156,10 @@ var _ = ginkgo.Describe(testsKey, func() {
 			testIsRedHatRelease(containersUnderTest.oc)
 		}
 		testIsRedHatRelease(testOrchestrator.oc)
+
+		for _, containersUnderTest := range containersUnderTest {
+			testNamespace(containersUnderTest.oc)
+		}
 	}
 })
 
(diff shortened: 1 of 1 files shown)


test-network-function/generic/suite.go
```diff
@@ -242,3 +246,14 @@ func GetTestConfiguration() *tnfConfig.TestConfiguration {
...
 	return config
 }
+
+func testNamespace(oc *interactive.Oc) {
+	pod := oc.GetPodName()
+	container := oc.GetPodContainerName()
+	ginkgo.When(fmt.Sprintf("Reading namespace of %s/%s", pod, container), func() {
+		ginkgo.It("Should not be 'default' and should not begin with 'openshift-'", func() {
+			gomega.Expect(oc.GetPodNamespace()).To(gomega.Not(gomega.Equal("default")))
+			gomega.Expect(oc.GetPodNamespace()).To(gomega.Not(gomega.HavePrefix("openshift-")))
```
ryandgoulding, line 256: All of this is just a reminder to myself, and no changes are actually required.

This is an interesting chicken/egg problem with our current configuration methodology.  `oc.GetPodNamespace()` retrieves the configuration value for namespace, which is required to set up the interactive `oc` session.  In other words, the value is directly configured by the user in yaml prior to running the suite.  Thus, this is testing the configuration, and not the active state of the cluster.

However, the command to derive a Pod namespace actually requires the namespace.  I.e.,
`oc get pod <name> -n <namespace> -o json | jq ".metadata.namespace"`

I think for now, we can assume that if we have gotten this far in the test suite, that the namespaces must be correct, as the interactive Contexts are setup and working.  Additionally, this information will eventually be polled using a label or some other mechanism as a part of simplifying the test configuration, and that can be the point where namespaces are sourced from the environment.

ryandgoulding (approved):

PR #129, opened 2021-03-26
Add autodiscovery for local kubeconfig files

mkochanowski:
Simplifies the `run-container.sh` script's interface by allowing the user to omit specifying the path to the kubeconfig file.
The script will perform actions to autodiscover an existing configuration.

These changes make the `-k` parameter optional.
Including the autodiscovery process, the local kubeconfig lookup order is as follows:
1. If -k is specified, use the paths provided with the -k option.
2. If -k is not specified, use paths defined in `$KUBECONFIG` on the underlying host.
3. If no paths are defined, use the default kubeconfig file located in `$HOME/.kube/config`.

To address the common case of using multiple kubeconfigs in the `$KUBECONFIG` environment variable, support for supplying more than one kubeconfig file has been added.
Each provided local kubeconfig file is mounted to the TNF container as a separate, read-only volume.

To maintain backward compatiblity with the existing TNF container image, indexing of container-side kubeconfigs starts from the second file, e.g.
- /usr/tnf/kubeconfig/config
- /usr/tnf/kubeconfig/config.2
- /usr/tnf/kubeconfig/config.3
- [...]

If multiple kubeconfigs are to be mounted to the container, a new `$KUBECONFIG` environment variable must be constructed and passed to the TNF container.
This variable must contain all paths to the mounted kubeconfig files, separated by a colon, for example:
`KUBECONFIG=/usr/tnf/kubeconfig/config:/usr/tnf/kubeconfig/config.2`

Both the script's usage prompt and README have been updated to reflect the new changes.

### DIFF
--- a/run-container.sh
+++ b/run-container.sh
@@ -1,22 +1,66 @@
 #!/usr/bin/env bash
 
 # Requires
-# kube config file mounted to /usr/tnf/kubeconfig/config
-# TNF config files mounted into /usr/tnf/config
-# A directory to output claim into mounted at /usr/tnf/claim
+# - kubeconfig file mounted to /usr/tnf/kubeconfig/config
+#   If more than one kubeconfig needs to be used, bind
+#   additional volumes for each kubeconfig, e.g.
+#     - /usr/tnf/kubeconfig/config
+#     - /usr/tnf/kubeconfig/config.2
+#     - /usr/tnf/kubeconfig/config.3
+# - TNF config files mounted into /usr/tnf/config
+# - A directory to output claim into mounted at /usr/tnf/claim
+# - A $KUBECONFIG environment variable passed to the TNF container
+#   containing all paths to kubeconfigs located in the container, e.g.
+#   KUBECONFIG=/usr/tnf/kubeconfig/config:/usr/tnf/kubeconfig/config.2
 
+REQUIRED_VARS=('LOCAL_KUBECONFIG' 'LOCAL_TNF_CONFIG' 'OUTPUT_LOC')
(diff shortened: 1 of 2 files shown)


run-container.sh
```diff
@@ -1,42 +1,151 @@
...
+			echo "$0: error: $error_message" 1>&2
+			var_missing=true
+		fi
+	done
+
+	if $var_missing; then
+		echo ""
+		usage_error
+	fi
+}
```
mkochanowski, line 93: Instead of differentianting between _required_ and _optional_ CLI parameters, I chose to check for the existence of variables required to run the TNF container.
This seems to me to be a much simpler approach, especially when dealing with autodiscovery for some of our variables.

I've left in the naive checking for the required number of arguments.

run-container.sh
```diff
@@ -1,42 +1,151 @@
...
+	# For example:
+	# - /usr/tnf/kubeconfig/config
+	# - /usr/tnf/kubeconfig/config.2
+	# - /usr/tnf/kubeconfig/config.3
+	if (($local_path_index > 0)); then
+		kubeconfig_index=$(($local_path_index + 1))
+		kubeconfig_path="$kubeconfig_path.$kubeconfig_index"
+	fi
+	echo $kubeconfig_path
+}
```
mkochanowski, line 126: Different naming schemes can be explored here, we could try to preserve some part of the original path/filename of each kubeconfig fetched from the host, it could make debugging configuration issues easier.

Not sure if it would be beneficial, I find the script to be quite transparent about what files will be mounted, and which names correspond to what.

crwr45 (approved):

PR #148, opened 2021-04-07
Test NodePort services do not exist

ydayagi:
We want to make sure that CNF does not use services of type NodePort

https://issues.redhat.com/browse/CTONET-781

### DIFF
--- a/pkg/tnf/handlers/nodeport/nodeport.go
+++ b/pkg/tnf/handlers/nodeport/nodeport.go
@@ -1,0 +1,99 @@
+// Copyright (C) 2021 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package nodeport
+
+import (
+	"strings"
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	npRegex = "(?s).+"
+)
+
+type NodePort struct {
+	result  int
+	timeout time.Duration
+	args    []string
+}
+
+func NewNodePort(timeout time.Duration, podNamespace string) *NodePort {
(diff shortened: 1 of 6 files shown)


pkg/tnf/handlers/nodeport/nodeport.go
```diff
@@ -0,0 +1,100 @@
...
+	result  int
+	timeout time.Duration
+	args    []string
+}
+
+func NewNodePort(timeout time.Duration, podNamespace string) *NodePort {
+	return &NodePort{
+		timeout: timeout,
+		result:  tnf.ERROR,
+		args: []string{"oc -n " +
```
ryandgoulding, line 42: This should be an array of elements separated by space.  I.e.,
```
args: []string{"oc", "-n", ...}
```

pkg/tnf/handlers/nodeport/nodeport.go
```diff
@@ -0,0 +1,100 @@
...
+// ReelFirst returns a step which expects the ping statistics within the test timeout.
+func (np *NodePort) ReelFirst() *reel.Step {
+	return &reel.Step{
+		Expect:  []string{npRegex},
+		Timeout: np.timeout,
+	}
+}
+
+func (np *NodePort) ReelMatch(_, _, match string) *reel.Step {
+	numExpectedLines := 1
```
ryandgoulding, line 77: I assume that the one expected line is just the header:
```
NAME         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
```
Might be helpful to add a comment if that is the case, in case someone has to debug this later :).

pkg/tnf/identifier/identifiers.go
```diff
@@ -195,6 +196,18 @@ var Catalog = map[string]TestCatalogEntry{
...
+	nodePortIdentifierURL: {
+		Identifier:  NodePortIdentifier,
+		Description: "A generic test used to test services of CNF pod's namespace.",
+		Type:        Normative,
+		IntrusionSettings: IntrusionSettings{
+			ModifiesSystem:           false,
+			ModificationIsPersistent: false,
+		},
+		BinaryDependencies: []string{
+			dependencies.OcBinaryName,
```
ryandgoulding, line 208: Please add `grep` here:
https://github.com/test-network-function/test-network-function/blob/main/pkg/tnf/dependencies/binaries.go
And then reference in the `BinaryDependencies` array here.

pkg/tnf/handlers/nodeport/nodeport.go
ydayagi: I think it is more readable this way. Also, there's no benefit when using an array and not a single string. All it does is add surrounding spaces.

pkg/tnf/identifier/identifiers.go
ydayagi: will do

pkg/tnf/handlers/nodeport/nodeport.go
ryandgoulding: I disagree.  No value in your context, but the `tnf.Tester` interface could be reused at some point, which is the whole point of using an interface in the first place:

https://github.com/test-network-function/test-network-function/blob/main/pkg/tnf/test.go#L45-L46

I will submit a patch clarifying that tnf.Tester interface's `args` is expected to contain the space separated elements of a command.

PR #150, opened 2021-04-07
Resolves workflow dispatch for force-push situations

ryandgoulding:


ryandgoulding:
This should not have passed the fixup commit.

PR #157, opened 2021-04-12
Test that node kernels are not tainted

ydayagi:
https://issues.redhat.com/browse/CTONET-844

### DIFF
--- a/pkg/tnf/identifier/identifiers.go
+++ b/pkg/tnf/identifier/identifiers.go
@@ -32,6 +32,8 @@ const (
 	roleBindingIdentifierURL        = "http://test-network-function.com/tests/rolebinding"
 	clusterRoleBindingIdentifierURL = "http://test-network-function.com/tests/clusterrolebinding"
 	nodePortIdentifierURL           = "http://test-network-function.com/tests/nodeport"
+	nodeNamesIdentifierURL          = "http://test-network-function.com/tests/nodenames"
+	nodeTaintedIdentifierURL        = "http://test-network-function.com/tests/nodetainted"
 
 	versionOne = "v1.0.0"
 )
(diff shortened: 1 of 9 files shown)


pkg/tnf/identifier/identifiers.go
```diff
@@ -209,6 +211,31 @@ var Catalog = map[string]TestCatalogEntry{
...
+	nodeTaintedIdentifierURL: {
+		Identifier:  NodeTaintedIdentifier,
+		Description: "A generic test used to test whether node is tainted",
+		Type:        Normative,
+		IntrusionSettings: IntrusionSettings{
+			ModifiesSystem:           false,
+			ModificationIsPersistent: false,
+		},
+		BinaryDependencies: []string{
+			dependencies.OcBinaryName,
```
ryandgoulding, line 235: Also requires `echo`.

ryandgoulding (changes requested):

ryandgoulding:
The only reservation I have about merging this involves the fact that doing so with this in the `generic` test suite will cause minikube `local-test-infra` to fail.  This is due to the fact that opening a debug pod for the `minikube` node does not work.  We will need some sort of flag, environment variable, etc. to further restrict the tests aside from `ginkgo.focus` so that we can skip this locally.  Need a few minutes to think about the best way to handle this going forward.

ydayagi:
@ryandgoulding what can be done to make this PR pass?

ryandgoulding:
@ydayagi can you make this test optional using based one whether the `TNF_MINIKUBE_ONLY` is set to `true` or `false`?  Aside from that, please just resolve the conflicts.

ryandgoulding (approved):

PR #128, opened 2021-03-24
No packages installed at runtime test

isaacdorfman:

PR #159, opened 2021-04-13
Combine the config mechanisms

crwr45:
Previously the "generic" and "multus" specs were in one config file
and "container" and "operator" specs were in a different file.

This change combines the two into a single config file that is
loaded once and then available to any spec on-demand. The location
of this config file can be set with an environment variable.

This also means that the configpool is no longer needed as a
separate item.

As part of this change set the complex keys from the "generic"
config file were removed to improve compatibility with JSON, and
remove code. The data is now in a list of maps. The removal of
this natural key does not impact performance as the first thing
done in the "generic" spec is that a new map is built.


README.md
```diff
@@ -207,19 +207,23 @@ for a future version.
 
 ## Test Configuration
 
-There are a few pieces of configuration required to allow the test framework to access and test the CNF:
+Configuration is accomplished with `tnf_config.yml` by default. An alternative configuration can be provided using the
+`TEST_CONFIGURATION_PATH` environment variable.
```
ryandgoulding, line 211: Should we adhere to prefixing all Environment variables with `TNF_`?

README.md
```diff
@@ -235,21 +239,18 @@ orchestrator.
...
-`cnf_test_configuration.yml` defines the roles under which operators and containers are to be tested.
+The `cnfs` and `operators` sections define the roles under which operators and containers are to be tested.
 
-[The included example config](test-network-function/cnf_test_configuration.yml) is set up with some examples of this:
+[The included example config](test-network-function/tnf_config.yml) is set up with some examples of this:
 It will run the `"OPERATOR_STATUS"` tests (as defined in `testconfigure.yml`) against an etcd operator, and the
 `"PRIVILEGED_POD"` and `"PRIVILEGED_ROLE"` tests against an nginx container.
 
-A more extensive example is provided in [pkg/config/example.yaml](pkg/config/example.yaml)
+A more extensive example of these sections is provided in [pkg/config/example.yaml](pkg/config/cnf_example.yaml)
```
ryandgoulding, line 253: Maybe we ought to move this into the `examples` directory at the root of the project.  That way it is dis-aggregated from regular source code.

pkg/config/config.go
```diff
@@ -16,12 +16,79 @@
...
+	"io/ioutil"
+	"os"
+
+	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
+)
 
 const (
-	filenameSuffix = "_test_configuration"
+	configurationFilePathEnvironmentVariableKey = "TEST_CONFIGURATION_PATH"
```
ryandgoulding, line 29: The more I think about it (and I realize this is from my submission originally), we should prefix with `TNF_`.

pkg/config/config.go
```diff
@@ -16,12 +16,79 @@
...
+	}
+	loaded = true
+	return nil
+}
+
+// GetConfigInstance provides access to the singleton ConfigFile instance.
+func GetConfigInstance() File {
+	if !loaded {
+		filePath := getConfigurationFilePathFromEnvironment()
+		log.Debugf("getConfigInstance before config loaded, loading from file: %s", filePath)
```
ryandgoulding, line 87: Capitalize `GetConfigInstance`.

pkg/config/container_test.go
```diff
@@ -74,14 +73,54 @@ const (
 	testNameSpace = "default"
 )
 
+const (
+	filePerm = 0644
```
ryandgoulding, line 77: Can you add a comment about what this is used for?

pkg/config/generic_config.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (C) 2020 Red Hat, Inc.
```
ryandgoulding, line 1: I think the correct way to deal with this is to use `2020-2021` now.

pkg/config/testdata/tnf_test_config.yml
```diff
@@ -0,0 +1,43 @@
...
+    containerName: partner
+operators:
+  - name: etcdoperator.v0.9.4
+    namespace: default
+    autogenerate: false
+    tests:
+      - OPERATOR_STATUS
+    certifiedoperatorrequestinfo:
+      - name: "etcd-operator"
+        organization: "redhat-marketplace"
```
ryandgoulding, line 34: I want to ensure that this is not actually reaching out to the API from the unit test context?  If so, then if that service falls over, our unit tests will perpetually fail.

test-network-function/tnf_config.yml
```diff
@@ -0,0 +1,43 @@
+generic:
+  containersUnderTest:
+    - namespace: default
```
ryandgoulding, line 3: Should be `tnf`.

test-network-function/tnf_config.yml
```diff
@@ -0,0 +1,43 @@
...
+      containerName: test
+      defaultNetworkDevice: eth0
+      multusIpAddresses:
+        - 10.217.0.8
+    - namespace: default
+      podName: partner
+      containerName: partner
+      defaultNetworkDevice: eth0
+      multusIpAddresses:
+        - 10.217.0.29
```
ryandgoulding, line 14: `tnf/partner/partner` is not actually under test, and should not be a container in this list IMO.

test-network-function/tnf_config.yml
```diff
@@ -0,0 +1,43 @@
...
+  partnerContainers:
+    - namespace: default
+      podName: partner
+      containerName: partner
+      defaultNetworkDevice: eth0
+      multusIpAddresses:
+        - 10.217.0.29
```
ryandgoulding, line 21: We need to take this into account:
https://github.com/test-network-function/test-network-function/pull/128
Particularly:
https://github.com/test-network-function/test-network-function/blob/main/test-network-function/generic_test_configuration.yml#L19-L22

ryandgoulding (changes requested):

test-network-function/tnf_config.yml
crwr45: Missed that with rebasing, should be sorted now.

pkg/config/testdata/tnf_test_config.yml
crwr45: Not at all, this config is not used for anything except to load a file and check that (some) of it is loaded successfully.
I had exactly the same concern with the tests for that spec itself.

pkg/config/generic_config.go
crwr45: This should be a global change across the repo in a new PR. I'll do it separately.

ryandgoulding (approved):

PR #164, opened 2021-04-19
Test worker nodes' hugepages

ydayagi:
Verify that CNFs did not change the hugepages configuration.
We use MachineConfig for determining cluster configuration

https://issues.redhat.com/browse/CTONET-860

### DIFF
--- a/pkg/tnf/handlers/hugepages/hugepages.go
+++ b/pkg/tnf/handlers/hugepages/hugepages.go
@@ -1,0 +1,173 @@
+// Copyright (C) 2021 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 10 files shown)


pkg/tnf/handlers/hugepages/hugepages.go
```diff
@@ -0,0 +1,173 @@
...
+			if isHugepagesParam(name) {
+				params[name] = value
+			}
+		}
+	}
+
+	// Use params for determining the hugepages settings
+	hugepages, ok := params["hugepages"]
+	if ok {
+		hp.hugepages, _ = strconv.Atoi(hugepages)
```
ryandgoulding, line 112: Same here.  I am fine with this but just wanted to note in case Charlie takes a look.  The err can be ignored as the output is well known to be an integer.

ryandgoulding (approved):

PR #165, opened 2021-04-20
CTONET-645 - scc verification

pkliczewski:
In this test we check whether `clusterPermissions` are set for specific `ClusterServiceVersion` is installed. We test etcd operator which is not using the permissions but this tests enable us to check whether operator in test is using `privileged` and in this case we should fail and analyze why such permission is needed.

Signed-off-by: Piotr Kliczewski <piotr.kliczewski@gmail.com>

### DIFF
--- a/pkg/config/example.yaml
+++ b/pkg/config/example.yaml
@@ -28,6 +28,7 @@ operators:
     tests:
       - CSV_INSTALLED
       - SUBSCRIPTION_INSTALLED
+      - CSV_SCC
     certifiedoperatorrequestinfo:
       - name: "etcd-operator"
         organization: "redhat-marketplace"
--- a/README.md
+++ b/README.md
@@ -124,8 +124,9 @@ accessed in user space, the test is unidirectional.
 
 ### operator tests
 
-Currently, the `operator` test spec is limited to two test cases called `OPERATOR_STATUS`.  `OPERATOR_STATUS`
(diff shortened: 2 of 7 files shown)


pkg/config/example.yaml
```diff
@@ -28,6 +28,7 @@ operators:
     tests:
       - CSV_INSTALLED
       - SUBSCRIPTION_INSTALLED
+      - CSV_SCC
     certifiedoperatorrequestinfo:
       - name: "etcd-operator"
```
pkliczewski, line 33: We are testing `etcd-operator` which is not using cluster permissions should we start to add one more?

ryandgoulding: Hi @pkliczewski.  This is just the default config, and will be updated per CNF.  But I agree, absolutely, in the case of CI, we will want to add more test cases (both positive and negative).  That is tracked by https://issues.redhat.com/browse/CTONET-743.

We currently do not have anyone working on this CTONET-743, as Gary G. has been moved to a separate effort.  I have brought this up with Jim E., and we don't have a clear solution going forward quite yet.  The problem is that this sort of CI/CD pipeline (Zuul or otherwise) requires dedicated hardware (or VMs), and goes above and beyond what our free GitHub Actions can support in terms of sanity tests.

If CTONET-743 interests you, or would be a good fit for someone else on the team, please feel free to take a look.  We are in the nascent stages of providing any smoke tests at all, but in the future minikube will certainly not suffice.  We require a robust OpenShift installation for true CI/CD.

ryandgoulding (approved):

PR #169, opened 2021-04-27
Provide ability to define Go-templated JSON tests

ryandgoulding:
Go Template Language is ubiquitous within Go programming environments.
Although there are some better alternatives, including Jinja2, Go Template
language is the obvious first choice for generating JSON tests from Go
templates.  The inclusion of such API(s) does not prohibit supporting other
types of templates in the future.

This change includes the ability to template JSON-based tests using Go template
language.  This is particularly useful for commands that require dynamic input,
such as `ping` destination.  There are several other examples, particularly for
plumbing input into commands.

Examples are added to the "examples/generic/template" directory.  Additionally,
use of these new API(s) is documented in DEVELOPING.md.


pkg/tnf/handlers/generic/generic.go
```diff
@@ -17,11 +17,15 @@
 package generic
 
 import (
+	"bytes"
 	"encoding/json"
+	"html/template"
```
crwr45, line 22: I assume the reason for using HTML templating is because it has escaping, and the input may end up being executed on a command line. I couldn't see anything in the `template` library docs about this use case but it seems sensible.

crwr45 (approved):

pkg/tnf/handlers/generic/generic.go
ryandgoulding: This was originally chosen for security reasons.  However, this may lead to adverse side effects.  I will go ahead and switch to `text/template`.  In reality, our code is executing (sometimes root) commands all over the place, and is not designed to be security software.

pkg/tnf/handlers/generic/testdata/bad_yaml.yaml
```diff
@@ -0,0 +1 @@
+Not YAML!!!!
```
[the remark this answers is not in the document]
ryandgoulding: Hm... you may be correct here.  I will update the contents of this file to something like `Not a key value pair!!!`

PR #172, opened 2021-04-28
removing custom ginkgo reporters

wying3:
Ginkgo has added junit support to the core so the custom
reporters are no longer needed and can be removed.

### DIFF
--- a/run-cnf-suites.sh
+++ b/run-cnf-suites.sh
@@ -36,6 +36,9 @@ done
 GINKGO_ARGS="-ginkgo.v -junit . -report $OUTPUT_LOC -claimloc $OUTPUT_LOC"
 FOCUS=""
 
+# specify Junit report file name.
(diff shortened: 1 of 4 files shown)


run-cnf-suites.sh
```diff
@@ -36,6 +36,9 @@ done
 GINKGO_ARGS="-ginkgo.v -junit . -report $OUTPUT_LOC -claimloc $OUTPUT_LOC"
 FOCUS=""
 
+# specify Junit report file name.
+GINKGO_ARGS2="-ginkgo.reportFile $OUTPUT_LOC/"cnf-certification-tests_junit.xml""
```
crwr45, line 40: Please add this to the existing `GINKGO_ARGS` declaration rather than specify a new one.

crwr45 (changes requested):

crwr45 (approved):

PR #173, opened 2021-04-28
ensure a CNF is usable after cordoning a node

ydayagi:
Read the CNF namespace's deployments
Drain at least one node that is executing deployment pods

https://issues.redhat.com/browse/CTONET-877

### DIFF
--- a/pkg/tnf/handlers/deployments/deployments.go
+++ b/pkg/tnf/handlers/deployments/deployments.go
@@ -1,0 +1,140 @@
+// Copyright (C) 2021 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package deployments
+
+import (
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	dpRegex = "(?s).+"
+)
+
+// Deployment holds information about a single deployment
+type Deployment struct {
+	Replicas    int
+	Ready       int
+	UpToDate    int
+	Available   int
+	Unavailable int
+}
+
+// DeploymentMap name to Deployment
+type DeploymentMap map[string]Deployment
+
+// Deployments holds information derived from running "oc -n <namespace> get deployments" on the command line.
+type Deployments struct {
+	deployments DeploymentMap
+	result      int
+	timeout     time.Duration
+	args        []string
+}
+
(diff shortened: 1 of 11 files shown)


ydayagi:
@ryandgoulding I created this draft PR so we can review the code and idea before end of impl

pkg/tnf/handlers/deploymentsdrain/deploymentsdrain.go
```diff
@@ -0,0 +1,95 @@
...
+// DeploymentsDrain holds information derived from running "oc adm drain" on the command line.
+type DeploymentsDrain struct {
+	result  int
+	timeout time.Duration
+	args    []string
+}
+
+// NewDeploymentsDrain creates a new DeploymentsDrain tnf.Test.
+func NewDeploymentsDrain(timeout time.Duration, node string) *DeploymentsDrain {
+	drainTimeout := timeout * drainTimeoutPercentage / 100
```
ryandgoulding, line 41: Just in case others are reviewing and wondering;  it would appear that time.Duration is smart enough to handle `float-like` numbers and possible type-upgrades even when multiplying by `int` types.  This is the type of Go operator overloading behavior I am not super familiar with, and had to punch it into the playground to confirm:

https://play.golang.org/p/2bMweG0-5jZ

pkg/tnf/handlers/deploymentsnodes/deploymentsnodes.go
```diff
@@ -0,0 +1,150 @@
...
+	"strings"
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	dnRegex      = "(?s).+"
```
ryandgoulding, line 30: Same as above;  can we use something similar to:
```
(?m)(NAME\s+)(NODE)(.*)
```
This would avoid matching the empty string, and hopefully any race conditions as well.

ryandgoulding (changes requested):

pkg/tnf/handlers/deployments/deployments.go
```diff
@@ -0,0 +1,140 @@
...
+	"strings"
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	dpRegex = "(?s).+"
```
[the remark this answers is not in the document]
ydayagi: This regex is already in use in previous handlers.
It does not allow empty output.
It expects at least one character.
This is how I distinguish error from successful execution.
By successful I mean that the executable was executed and produced some kind of output.

pkg/tnf/handlers/deploymentsdrain/deploymentsdrain.go
ydayagi: I am not sure what you're trying to focus on here.
If you are referring to the Duration type's ability to hold float then the answer is yes.
I read docs and tested.

pkg/tnf/handlers/deploymentsnodes/deploymentsnodes.go
ydayagi: see previous comments.

pkg/tnf/handlers/deployments/deployments.go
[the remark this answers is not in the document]
ryandgoulding: Oh okay, I misread the `+` as a `*`.  You are right, this is fine.

pkg/tnf/handlers/deploymentsdrain/deploymentsdrain.go
ryandgoulding: Yes, just leaving it in case Charlie or anyone else reviewing was curious.  No action is needed.

ryandgoulding (approved):

PR #178, opened 2021-05-03
CTONET-845 - added test for machineconfig and node kernel params comp…

isaacdorfman:
…atibility

### DIFF
--- a/pkg/tnf/handlers/currentkernelcmdlineargs/currentkernelcmdlineargs.go
+++ b/pkg/tnf/handlers/currentkernelcmdlineargs/currentkernelcmdlineargs.go
@@ -1,0 +1,96 @@
+// Copyright (C) 2021 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package currentkernelcmdlineargs
+
+import (
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/dependencies"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	successfulOutputRegex = `.+`
+)
+
+// CurrentKernelCmdlineArgs holds information regarding /proc/cmdline.
+type CurrentKernelCmdlineArgs struct {
+	kernelArguments string // Output variable that stores the kernel arguments of the mc
+	result          int
+	timeout         time.Duration
+	args            []string
+}
+
+// NewCurrentKernelCmdlineArgs creates a CurrentKernelCmdlineArgs tnf.Test.
+func NewCurrentKernelCmdlineArgs(timeout time.Duration) *CurrentKernelCmdlineArgs {
+	return &CurrentKernelCmdlineArgs{
+		timeout: timeout,
+		result:  tnf.ERROR,
+		args:    []string{dependencies.CatBinaryName, "/proc/cmdline"},
+	}
+}
+
+// GetKernelArguments returns the kernel arguments extracted from /proc/cmdline while running the CurrentKernelCmdlineArgs tnf.Test.
+func (cmdlineArgs *CurrentKernelCmdlineArgs) GetKernelArguments() string {
+	return cmdlineArgs.kernelArguments
+}
+
+// Args returns the command line args for the test.
+func (cmdlineArgs *CurrentKernelCmdlineArgs) Args() []string {
+	return cmdlineArgs.args
+}
+
+// GetIdentifier returns the tnf.Test specific identifier.
+func (cmdlineArgs *CurrentKernelCmdlineArgs) GetIdentifier() identifier.Identifier {
+	return identifier.CurrentKernelCmdlineArgsURLIdentifier
+}
+
+// Timeout returns the timeout in seconds for the test.
+func (cmdlineArgs *CurrentKernelCmdlineArgs) Timeout() time.Duration {
+	return cmdlineArgs.timeout
+}
+
+// Result returns the test result.
+func (cmdlineArgs *CurrentKernelCmdlineArgs) Result() int {
+	return cmdlineArgs.result
+}
+
+// ReelFirst returns a step which expects the /proc/cmdline kernel arguments within the test timeout.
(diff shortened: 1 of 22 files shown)


pkg/tnf/handlers/bootconfig/bootconfig.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	successfulOutputRegex = `(?s).+`
```
ryandgoulding, line 28: One limitation of using an expect based process to control the PTY is the fact that we shouldn't trigger `ReelMatch(...)` based on empty (nil) output. The reason for this is that, if you match the empty string, `ReelMatch(...)` is triggered regardless of whether command execution has completed, which can lead to some nasty and inconsistent race conditions.

As of now, we do not mandate this requirement in code, but we should. Can you instead change to a regular expression that expects some output?

Additionally, I will create a story to address checking that a given reel.Handler does not contain an reel.Step.Expect regular expression string that matches the empty String. Unfortunately, since Go does not support constructors (or inheritance really well for that matter), this will need to be a runtime check at the time of the test Run; better than nothing though.

pkg/tnf/handlers/bootconfig/bootconfig.go
```diff
@@ -0,0 +1,97 @@
...
+	args       []string
+}
+
+// NewBootConfig creates a BootConfig tnf.Test.
+func NewBootConfig(timeout time.Duration, mcName string) *BootConfig {
+	return &BootConfig{
+		timeout: timeout,
+		result:  tnf.ERROR,
+		args: []string{
+			"echo \"cat /host/boot/loader/entries/\\`ls -l /host/boot/loader/entries/ | sort -r -k 9 | head -n 1 | cut -d\\\" \\\" -f9 -\\`\" | oc debug -q node/crc-l6qvn-master-0",
```
ryandgoulding, line 45: I am a bit confused by what this actually does.  It looks like it is hard-coded for a specific CRC `Node` though, and will not work against a stock cluster.  Not every cluster will have `node/crc-l6qvn-master-0` installed.

Can you provide a way to make the node name configurable?

pkg/tnf/handlers/currentkernelcmdlineargs/currentkernelcmdlineargs.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// NewCurrentKernelCmdlineArgs creates a CurrentKernelCmdlineArgs tnf.Test.
+func NewCurrentKernelCmdlineArgs(timeout time.Duration, mcName string) *CurrentKernelCmdlineArgs {
+	return &CurrentKernelCmdlineArgs{
+		timeout: timeout,
+		result:  tnf.ERROR,
+		args: []string{
+			"cat /proc/cmdline",
+		},
```
ryandgoulding, line 46: Please use an actual `[]string{}` string array.  Even though this currently works, the API expects a String array.  So something like this:
```
args: []string{dependencies.CatBinaryName, "/proc/cmdline"}
```

This will make it easier to eventually change this interface, as it is currently bastardized in many places.

pkg/tnf/handlers/mckernelarguments/mckernelarguments.go
```diff
@@ -0,0 +1,97 @@
...
+	args            []string
+}
+
+// NewMcKernelArguments creates a NodeMcName tnf.Test.
+func NewMcKernelArguments(timeout time.Duration, mcName string) *McKernelArguments {
+	return &McKernelArguments{
+		timeout: timeout,
+		result:  tnf.ERROR,
+		args: []string{
+			"oc get mc " + mcName + " -o jsonpath=\"{.spec.kernelArguments}\"",
```
ryandgoulding, line 45: Similar to above, please expand the string array.  So, this should be:
```
args: []string{
dependencies.OcBinaryName,
"get", "mc", mcName", "-o", "jsonpath=\"{.spec.kernelArguments}\""
}

**Comment by ryandgoulding** [comment]
Each added `reel.Handler` requires unit tests before this PR can be considered for merge.  For an example, please look to [ping_test.go|https://github.com/test-network-function/test-network-function/blob/main/pkg/tnf/handlers/ping/ping_test.go].

**Review by ryandgoulding — CHANGES_REQUESTED** [review]

File: test-network-function/generic/suite.go [hunk 5]
```diff
@@ -346,6 +358,116 @@ func testGracePeriod(context *interactive.Context, podName, podNamespace string)
...
+		if len(splitArgs) == 1 {
+			retval[splitArgs[0]] = ""
+		} else {
+			retval[splitArgs[0]] = splitArgs[1]
+		}
+	}
+	return retval
+}
+
+func filterArray(vs []string, f func(string) bool) []string {
```
crwr45, line 461: `filterArray` and `argListToMap` are at a very different level of abstraction to the rest of the file and as such feel out of place where they are. Please move them to another file, and add a small comment explaining their purpose.

crwr45 (changes requested):

pkg/tnf/handlers/currentkernelcmdlineargs/currentkernelcmdlineargs.go
isaacdorfman: Now it's fixed

pkg/tnf/handlers/mckernelarguments/mckernelarguments.go
isaacdorfman: Now it's fixed

test-network-function/generic/suite.go
isaacdorfman: Moved to "pkg/utils/utils.go"

pkg/tnf/handlers/bootconfig/bootconfig.go
```diff
@@ -0,0 +1,97 @@
...
+	args       []string
+}
+
+// NewBootConfig creates a BootConfig tnf.Test.
+func NewBootConfig(timeout time.Duration, nodeName string) *BootConfig {
+	return &BootConfig{
+		timeout: timeout,
+		result:  tnf.ERROR,
+		args: []string{
+			"echo \"cat /host/boot/loader/entries/\\`ls -l /host/boot/loader/entries/ | sort -r -k 9 | head -n 1 | cut -d\\\" \\\" -f9 -\\`\" | oc debug -q node/" + nodeName,
```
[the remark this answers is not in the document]
isaacdorfman: @crwr45 I don't understand the question.
And do you have an alternative method of doing this(without running oc debug twice)?
This script essentially checks the "/host/boot/loader/entries/" folder for the file with the highest id and reads it.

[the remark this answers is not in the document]
isaacdorfman: @crwr45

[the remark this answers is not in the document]
crwr45: As I said, this is more than I'm comfortable with in a single command. It's hard to understand, both due to the number of steps and the multiple levels of string escaping. It's also going to be hard to debug especially since, as I said above, the regex and code doesn't distinguish between success or errors.
My concern is primarily the readability and debugability of the code with so much complexity embedded in the `BootConfig` command string. If the way to improve this is to have more than one `oc debug` then please consider it, when they are at odds I think readability is more important than brevity.
Another option is to do more of this parsing in Go code, making it more testable.

crwr45 (approved):

ryandgoulding (approved):

PR #181, opened 2021-05-04
Pods should be managed by DaemonSets/Deployments

ydayagi:
Verify CNF pods' owners include Deployment or DaemonSet
https://issues.redhat.com/browse/CTONET-884

### DIFF
--- a/test-network-function/generic/suite.go
+++ b/test-network-function/generic/suite.go
@@ -43,6 +43,7 @@ import (
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/nodenames"
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/nodeport"
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/nodetainted"
+	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/owners"
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/ping"
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/rolebinding"
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/serviceaccount"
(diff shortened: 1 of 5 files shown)


test-network-function/generic/suite.go
```diff
@@ -208,6 +209,10 @@ var _ = ginkgo.Describe(testsKey, func() {
 		for _, containersUnderTest := range containersUnderTest {
 			testDeployments(containersUnderTest.oc.GetPodNamespace())
 		}
+
+		for _, containersUnderTest := range containersUnderTest {
+			testOwner(containersUnderTest.oc.GetPodNamespace(), containersUnderTest.oc.GetPodName())
```
crwr45, line 214: Out of interest, what's the reason for not using the existing `oc` context for the container?

test-network-function/generic/suite.go
```diff
@@ -208,6 +209,10 @@ var _ = ginkgo.Describe(testsKey, func() {
 		for _, containersUnderTest := range containersUnderTest {
 			testDeployments(containersUnderTest.oc.GetPodNamespace())
 		}
+
+		for _, containersUnderTest := range containersUnderTest {
```
crwr45, line 213: I know there's several of these loops which are reusing `containersUnderTest` to refer to a single container inside the loop but please don't add another one. We'll clean up the others separately at a later time.

crwr45 (changes requested):
One readability change, and one question. Otherwise this is great.

test-network-function/generic/suite.go
ydayagi: I need admin access to the cluster and I do not need to execute anything within the pod.

test-network-function/generic/suite.go
ydayagi: so merge into an existing loop?

ydayagi: I think I did not understand your first comment. What is the requested change?

crwr45: The term I was failing before to find was shadowing.
This `for` loop (and other similar ones in this file) is shadowing `containersUnderTest` inside the loop. This makes the code harder to understand and work with.
I'm not asking you to change all the occurrences of this (that should be done but will obfuscate the specifics of this change), rather not to add another one.

PR #183, opened 2021-05-04
export TNF_MINIKUBE_ONLY to true for k8s env testing.

wying3:
eventually export TNF_MINIKUBE_ONLY =true has to do manually.


README.md
```diff
@@ -86,6 +86,10 @@ The 'General' tests are designed to test any commodity CNF running on OpenShift,
 'CNF-specific' tests are designed to test some unique aspects of the CNF under test are behaving correctly.  This could
 include specifications such as issuing a `GET` request to a web server, or passing traffic through an IPSEC tunnel.
 
+## Turn off openshift required Tests when CNF run on kubernetes only environment
+when CNF test for CNF running on k8s only envirnment. execute commandline below on cli before compile tool and run shell script.
+   export TNF_MINIKUBE_ONLY=true
```
mkochanowski, line 91: I'd put the command in a [Fenced code block](https://github.github.com/gfm/#fenced-code-blocks), to make the documentation easier to digest.

crwr45, line 91: I'll second what Marek has said. You can also see the rest of the documentation to see how other examples are formatted.

crwr45 (changes requested):

crwr45 (changes requested):
Please `squash` your commits down into one, and `rebase` onto main to ensure the commits are linear.

crwr45:
Please [amend](https://docs.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message) the commit message to be cleaner.
A good rule of thumb is that a commit message has a short title on the top line that completes the sentence "if applied, this commit will ...", if more detail is needed then a blank line and more details. This change doesn't need the explanation.

Something like `git commit --amend -m "control FS Diff test with TNF_MINIKUBE_ONLY env var"`

test-network-function/generic/suite.go
```diff
@@ -164,7 +164,8 @@ var _ = ginkgo.Describe(testsKey, func() {
 		log.Info(containersUnderTest)
 
 		ginkgo.Context("Container does not have additional packages installed", func() {
-			if os.Getenv("FSDIFF") == "1" {
+			// use this boolean to turn off tests that require OS packages
+			if isMinikube() {
```
crwr45, line 168: This is the inverse of what's needed. We only want to run the test when the env var is false, this will only run it when the var is true.
Please make sure to check that the test runs when the env var is not set.

crwr45 (changes requested):

test-network-function/generic/suite.go
ryandgoulding: Switch to:
```
if !isMinikube() {
...
```

ryandgoulding (changes requested):

test-network-function/generic/suite.go
wying3: i think my last merge/rebase screwed the changes i did, will correct. thanks for point out.

PR #189, opened 2021-05-06
CTONET-645 - fix empty resourceNames

pkliczewski:
In case when resourceNames is not defined we returned empty string which failed the test. This PR checks whether there are no resource names and make sure we use well know token `EMPTY` to indicate this. The token was introduced to help validate whether path to

``` json
.spec.install.spec.clusterPermissions[].rules[].resourceNames
```

is valid. In order to use command introduced by this test in a test description we can use action deny with expectedstatus `privileged`.

Signed-off-by: Piotr Kliczewski <piotr.kliczewski@gmail.com>


crwr45 (changes requested):
Changes look sensible, just needs the long line wrapping to make it a bit more readable.

pkliczewski:
> Changes look sensible, just needs the long line wrapping to make it a bit more readable.

Ack

crwr45 (approved):

PR #192, opened 2021-05-10
Add support for external TNF source repositories

mkochanowski:
This change adds an optional build argument `TNF_SRC_URL` to the Dockerfile that allows to override the URL the TNF framework will be downloaded from during the image build process.
The value of the build argument points to the official TNF repository by default.

Also, the Dockerfile now supports checking out a commit of a TNF source repository, in addition to branches and tags.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN yum install -y golang-${GOLANG_VERSION} jq make git
 
 # Build oc from source
 ADD ${OC_SRC_URL} ${TEMP_DIR}
-RUN mkdir ${OC_SRC_DIR} && \ 
+RUN mkdir ${OC_SRC_DIR} && \
 	tar -xf ${TEMP_DIR}/${OC_SRC_ARCHIVE} -C ${OC_SRC_DIR} --strip-components=1 && \
 	cd ${OC_SRC_DIR} && \
 	make oc && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -33,8 +33,13 @@ ENV PATH="/root/go/bin:${PATH}"
...
 ARG TNF_VERSION
-# Pull the required version of TNF
-RUN git clone --depth=1 --branch=${TNF_VERSION} https://github.com/test-network-function/test-network-function ${TNF_SRC_DIR}
+ARG TNF_SRC_URL=https://github.com/test-network-function/test-network-function
+ARG GIT_CHECKOUT_TARGET=$TNF_VERSION
+
+# Clone the TNF source repository and checkout the target branch/tag/commit
+RUN git clone --no-single-branch --depth=1 ${TNF_SRC_URL} ${TNF_SRC_DIR}
+RUN git -C ${TNF_SRC_DIR} fetch origin ${GIT_CHECKOUT_TARGET}
+RUN git -C ${TNF_SRC_DIR} checkout ${GIT_CHECKOUT_TARGET}
```
mkochanowski, line 42: I've tried different approaches until I finally settled on this one.

One alternative solution would be to create a boolean build argument called something like `TNF_VERSION_AS_COMMIT_SHA`, that would then set which `git clone` operation would be performed (there would be a separate git clone syntax for branches/tags, and a separate for cloning by SHA of a commit). Sadly, introducing general conditional logic in Dockerfiles is nontrivial, and one-line conditional expressions significantly reduce readability of Dockerfile instructions.

I think the proposed approach is a nice compromise. Still quite clean, and does what it should.

PR #193, opened 2021-05-11
Add network mode settings to run-container.sh

mkochanowski:
This change adds an optional -n parameter to run-container.sh allowing to set network settings consumed by a `--network` parameter of the `docker run` command responsible for creating the TNF container.
The network mode defaults to `bridge`, consistent with the equivalent setting provided by the `docker run` command.

The documentation, and the usage prompt have been updated to explain the usage of the -n parameter.

Due to the increase in the number of options available, the list of optional arguments in the documentation is now sorted alphabetically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,9 @@ To pull the latest container and run the tests you use the following command. Th
 
 Optional arguments are:
 
-* `-k` gives a path to one or more kubeconfig files to be used by the container to authenticate with the cluster. Paths must be separated by a colon.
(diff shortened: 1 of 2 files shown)


crwr45 (commented):
I'm not 100% clear on why this change is needed, could you add a sentence or two to the commit on "why" for future reference? Otherwise, nice work. 

I think you sorted the options to avoid them spelling a word like `kin` or `ink`

PR #195, opened 2021-05-12
add extra info per test

ydayagi:
We want to be able to add informational messages to each test

https://issues.redhat.com/browse/CTONET-887

### DIFF
--- a/pkg/tnf/test.go
+++ b/pkg/tnf/test.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	expect "github.com/google/goexpect"
+	"github.com/onsi/ginkgo"
 	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
 	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
 )
@@ -32,6 +33,24 @@ const (
 	// FAILURE represents a failed test.
 	FAILURE
 )
+
+// TestsExtraInfo a collection of messages per test that is added to the claim file
+// use WriteTestExtraInfo for writing to it
+var TestsExtraInfo []map[string][]string = []map[string][]string{}
+
+// CreateTestExtraInfoWriter creates a function that writes info messages for a specific test
+// info messages that were already added by calling the function will exist in the claim file
+func CreateTestExtraInfoWriter() func(string) {
+	testName := ginkgo.CurrentGinkgoTestDescription().FullTestText
+	if testName == "" {
+		return func(string) {}
+	}
+	extraInfo := map[string][]string{testName: nil}
+	TestsExtraInfo = append(TestsExtraInfo, extraInfo)
+	return func(info string) {
+		extraInfo[testName] = append(extraInfo[testName], info)
+	}
+}
 
 // ExitCodeMap maps a test result value to a more appropriate Unix return code.
 var ExitCodeMap = map[int]int{
--- a/test-network-function/suite_test.go
+++ b/test-network-function/suite_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/test-network-function/test-network-function-claim/pkg/claim"
 	"github.com/test-network-function/test-network-function/pkg/config"
 	"github.com/test-network-function/test-network-function/pkg/junit"
+	"github.com/test-network-function/test-network-function/pkg/tnf"
 	_ "github.com/test-network-function/test-network-function/test-network-function/container"
 	"github.com/test-network-function/test-network-function/test-network-function/diagnostic"
 	_ "github.com/test-network-function/test-network-function/test-network-function/generic"
@@ -50,6 +51,7 @@ const (
 	reportFlagKey                 = "report"
 	// dateTimeFormatDirective is the directive used to format date/time according to ISO 8601.
 	dateTimeFormatDirective = "2006-01-02T15:04:05+00:00"
+	extraInfoKey            = "testsExtraInfo"
 )
 
 var (
@@ -115,6 +117,8 @@ func TestTest(t *testing.T) {
 	if err != nil {
 		log.Fatalf("error reading JUnit XML file into JSON: %v", err)
 	}
+
+	junitMap[extraInfoKey] = tnf.TestsExtraInfo
 
 	endTime := time.Now()
 	claimData.Results = junitMap
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -603,3 +603,35 @@ The same result could be achieved using a Go Template:
 ```shell-script
 oc get pod %s -n %s -o go-template='{{len .spec.containers}}{{"\n"}}'
 ```
+
+## Adding information to claim file
+
+The result of each test execution is included in the claim file.
+Sometimes it is convenient to add informational messages regarding the test execution.
+For this purpose we have an additional section in the claim file (see `suite.extraInfoKey` in suite_test.go).
+In order to add informational messages to your test use the function `tnf.CreateTestExtraInfoWriter`.
+This function adds an entry for your test in the claim file.
+The return value is a function.
+The returned function can be called to add a message to the test entry.
+Each added message will be written to claim file even if test failed or error occurred in the middle of the test.
+
+Example usage:
+```go
+ginkgo.It("Should do what I tell it to do", func(){
+  // do some work
+  // create test info writer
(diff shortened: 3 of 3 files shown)


pkg/tnf/test.go
```diff
@@ -33,6 +34,19 @@ const (
 	FAILURE
 )
 
+// TestsExtraInfo a collection of messages per test that is added to the claim file
+// use WriteTestExtraInfo for writing to it
+var TestsExtraInfo []map[string][]string = []map[string][]string{}
```
ryandgoulding, line 39: I believe we should have an opinion about the payload, rather than `[]string`.  Instead, can we use something like this as the payload:

```
type ExtraInfo struct {
  messages string
  timestamp Time
}
```

and consequently:

```
var TestsExtraInfo []map[string]ExtraInfo = []map[string]ExtraInfo{}
```

Instead of stacking messages, call `WriteTestExtraInfo(...)` for each invocation, so each message is timestamped, similar to the information we would get from `logrus`.

test-network-function/suite_test.go
```diff
@@ -116,6 +117,8 @@ func TestTest(t *testing.T) {
 		log.Fatalf("error reading JUnit XML file into JSON: %v", err)
 	}
 
+	junitMap["testsExtraInfo"] = tnf.TestsExtraInfo
```
ryandgoulding, line 120: Please use a constant for `"testsExtraInfo"`

pkg/tnf/test.go
```diff
@@ -33,6 +34,19 @@ const (
...
+
+// WriteTestExtraInfo adds a new object for the current test with the given info
+func WriteTestExtraInfo(info []string) {
```
crwr45, line 42: Please make this method able to add to an already existing entry in `TestsExtraInfo` for a test. As it is I think it's prone to unintentional misuse.

If `WriteTestExtraInfo` is called after any assertions have been made in a test it may never be called, and any information accumulated in the test up to that point (`extraInfo` in the example) will not be recorded.

crwr45 (changes requested):
Please make a way to add to an already existing entry in `TestsExtraInfo` for a test. As it is I think it's prone to unintentional misuse and loss of information. Specifically if `WriteTestExtraInfo` is called after any assertions have been made in a test it may never be called for that test, and any information accumulated in the test up to that point (`extraInfo` in the example) will not be recorded.

Given the point @ryandgoulding raised about key collisions this probably isn't just a case of checking for existence of a key and appending. I suggest a factory-type method in `test.go` that provides a test with a pre-configured method/struct can be used to log new messages for the test.

@ryandgoulding what do you think?

pkg/tnf/test.go
ydayagi: I do not see what value do u get from the timestamp. The 'problem' we see with the CNF is not related to a point in time and the messages are for general information and not logging.
There's a reason for providing a function that takes []string and not just string. I'll explain. We said that the test does not have a unique key. The ginkgo.FullSpecText can be the same for two instances of the same test. That's why we moved from a map of info to a slice of info. Ignore the current inner map. I'm referring to []map instead of map[]. Between calls to WriteExtraInfo, I do not know to which test to add the message. 2 options. One is letting the user supply a test object that he/she keeps locally through the test. Second, let the user store a slice locally and call WriteExtraInfo with the slice. I think option #2 is better for the user.

ydayagi:
@crwr45 @ryandgoulding regarding Charlie's comment. "factory-type method" is a bit vague. Please try to provide implementation details.
Also, please see my response to Ryan's comment. I think the two comments are related.
I think we're pushing towards a complex solution for such a simple feature.
Currently, the output is at least as good ad Ginkgo's report.
The user's effort is minimal and --> there's no solution that will make the user work less.
If all we want is to be able to track information for a test in the logger then all we need is to add a header/prefix to test messages. logrus supports that.

test-network-function/generic/suite.go
```diff
@@ -653,9 +653,13 @@ func getContext() *interactive.Context {
 func testOwner(podNamespace, podName string) {
 	ginkgo.When("Testing owners of CNF pod", func() {
 		ginkgo.It("Should contain at least one of kind DaemonSet/ReplicaSet", func() {
+			extraInfo := []string{}
 			context := getContext()
```
ryandgoulding, line 657: Can we add something like this in to resolve Charlie's point:
```
written := false
// emit informational message(s) in the case of early exit
defer func() {
    if !written {
        tnf.WriteTestExtraInfo(extraInfo)
    }
}()

context := getContext()
tester := owners.NewOwners(defaultTimeout, podNamespace, podName)
extraInfo = append(extraInfo, fmt.Sprintf("Pod %s/%s before execution", podNamespace, podName))
extraInfo = append(extraInfo, fmt.Sprintf("Pod %s/%s executed", podNamespace, podName))
tnf.WriteTestExtraInfo(extraInfo)
written = true
...
```

crwr45:
Ryans suggestion looks like a good solution that maintains simplicity on both sides.

For reference, I was imagining something like this entirely untested example, with a claim info recorder "factory" that returns a method configured to add messages to a certain place in `TestsExtraInfo`. It adds complexity on the implementation side, but not to the user, while ensuring each message is flushed at creation time.

test.go:
```
// NewClaimInfoRecorder will create and return a method that can be called to
// record extra information from a test. Each call to the returned method will add
// an item to the same index in the TestsExtraInfo list, under the same testName
// It is the callers responsibility to ensure that the method is not used outside the
// context in which it is meaningful (the test for which it was created)
func NewClaimInfoRecorder() (func(string)) {
    index := len(TestsExtraInfo)
    testName := ginkgo.CurrentGinkgoTestDescription().FullTestText
    TestsExtraInfo = append(TestsExtraInfo, map[string][]string{})
    return func(s string) {
        TestsExtraInfo[index][testName] = append(TestsExtraInfo[index][testName], s)
    }
}
```

suite.go
```
...
addInfoToClaim := NewClaimInfoRecorder()
tester := owners.NewOwners(defaultTimeout, podNamespace, podName)
addInfoToClaim(fmt.Sprintf("Pod %s/%s before execution", podNamespace, podName)
addInfoToClaim(fmt.Sprintf("Pod %s/%s executed", podNamespace, podName))
...
```

test-network-function/generic/suite.go
ydayagi: The defer is in the context of the test and not in the code I supply.

ryandgoulding: I disagree.  You are streaming a slice of messages in this example, and if an error occurs (such as in line 660), then the stream is never flushed, and all context is lost.

ryandgoulding (changes requested):

test-network-function/generic/suite.go
ryandgoulding: I am a bit confused by your question, so let me attempt to clarify.  You edited `test-network-function/generic/suite.go` in this Pull Request to include the call to the `WriteTestExtraInfo(...)` API added `pkg/tnf/test.go`.

I am advocating that the Client usage of `WriteTestExtraInfo(...)` in `suite.go` is done in a `defer func(){}` in order to ensure that any intermediate `extraInfo` is written to the claim, regardless of whether an exception occurs in the `Ginkgo.It` block started on line 655.  The `defer func(){}` will be called when the anonymous function on line 655 exits, regardless of whether there is an intermediate error.  Thus any in-flight `extraInfo` strings will be written to the claim.

ydayagi:
@ryandgoulding please see my latest changes. It solves all your comments.

ryandgoulding:
Are the changes staged?  https://github.com/test-network-function/test-network-function/commit/c4320cbe7658d209614f81670389a06d6c8e2d64 does not contain the `defer` logic I mentioned.

crwr45:
Thank you, this is looking much better. To ensure this is used going forward please add a sentence or two at an appropriate point in `DEVELOPING.md` to explain this feature, and/or example usage in an existing test.

ydayagi:
> Thank you, this is looking much better. To ensure this is used going forward please add a sentence or two at an appropriate point in `DEVELOPING.md` to explain this feature, and/or example usage in an existing test.

Done

crwr45 (approved):

ryandgoulding (approved):

PR #201, opened 2021-05-13
Add automatic publishing of TNF container images

mkochanowski:
This change adds a `tnf-image.yml` workflow that first fetches the version number of the latest relase from the generic version branch (e.g. `1.0.x`), builds the image, performs smoke tests, and then publishes the new container image to Quay.io.

The workflow can be triggered by publishing a new release, or by manually dispatching it via the `workflow_dispatch` feature built into GitHub.

To automatically apply security patches provided in UBI, an additional cron trigger will ensure that the workflow is run and the image is rebuilt once a day.

### DIFF
--- a/.github/workflows/tnf-image.yaml
+++ b/.github/workflows/tnf-image.yaml
@@ -1,0 +1,128 @@
+name: 'Publish the `test-network-function` image (latest release only)'
+on:
+  # Run the workflow when a new release gets published
+  release:
+    types: [published]
+  # Run the workflow every day at 5 am UTC (1 am EST, 7am CET)
+  # This is useful for keeping the image up-to-date with security
+  # patches provided in the UBI.
+  # Disclaimer: There is no guarantee that scheduled workflows will
+  # run at the predefined time, if at all. The delay is usually
+  # around 10-30 minutes.
+  schedule:
+    - cron: '0 5 * * *'
+  workflow_dispatch:
+defaults:
+  run:
+    shell: bash
+env:
+  REGISTRY: quay.io
+  IMAGE_NAME: testnetworkfunction/test-network-function
+  IMAGE_TAG: latest
+  IMAGE_CONTAINER_FILE_PATH: Dockerfile
+  CURRENT_VERSION_GENERIC_BRANCH: 1.0.x
+  TNF_MINIKUBE_ONLY: true
+  TNF_CONFIG_DIR: /tmp/tnf/config
+  TNF_OUTPUT_DIR: /tmp/tnf/output
+  TNF_SRC_URL: 'https://github.com/${{ github.repository }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/tnf-image.yaml
```diff
@@ -0,0 +1,128 @@
...
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.REGISTRY }}
+          # Use a Robot Account to authenticate against Quay.io
+          # https://docs.quay.io/glossary/robot-accounts.html
+          username: ${{ secrets.QUAY_ROBOT_USERNAME }}
+          password: ${{ secrets.QUAY_ROBOT_TOKEN }}
+
+      - name: Push the newly built image to Quay.io
+        run: docker push --all-tags ${REGISTRY}/${IMAGE_NAME}
```
mkochanowski, line 128: Initially, pushing the image was handled by a dedicated job. 
GitHub makes it hard to persist resources between jobs, so the built image had to be `docker saved`, uploaded as an artifact, and then downloaded and `docker loaded` in the next job. This is really cumbersome, and adds unnecessary boilerplate to the workflow defintion.

An alternative approach would be to rebuild the image inside the push job, but building the TNF image takes about 5 minutes, so doing it twice increases the total execution time significantly. I also wanted to make sure that the image pushed to Quay.io is the exact one that has passed the testing.

.github/workflows/tnf-image.yaml
```diff
@@ -0,0 +1,128 @@
...
+  release:
+    types: [published]
+  # Run the workflow every day at 5 am UTC (1 am EST, 7am CET)
+  # This is useful for keeping the image up-to-date with security
+  # patches provided in the UBI.
+  # Disclaimer: There is no guarantee that scheduled workflows will
+  # run at the predefined time, if at all. The delay is usually
+  # around 10-30 minutes.
+  schedule:
+    - cron: '0 5 * * *'
```
mkochanowski, line 13: This will result in the TNF container image being rebuilt and pushed to Quay.io once a day.
That means a new entry in the [Tag history](https://quay.io/repository/testnetworkfunction/test-network-function?tab=history) will be added every day.

If we want to avoid that, there is a REST API that Quay.io exposes we could use to fetch the information about results of a security scan for a given image. This could be queried in a separate workflow, that would then trigger `tnf-image.yaml` only when a rebuild is necessary.

I suggest we stick with the current approach (or we could remove the `schedule` trigger completely), and treat the suggested improvements to the _vulnerability discovery and purging_ process as a separate story for future investigation.

ryandgoulding: Sounds logical to me.

ryandgoulding (approved):

PR #204, opened 2021-05-17
Container autoconfig functionality

crwr45:
Companion change to local-test-infra https://github.com/test-network-function/cnf-certification-test-partner/pull/16


README.md
```diff
@@ -282,6 +282,24 @@ be seen in [cnf-certification-test-partner](https://github.com/test-network-func
 the first entry found with `"default"=true` is used. This annotation is automatically managed in OpenShift but may not
 be present in K8s.
 
+### Container Spec
+
+* Pods to be tested by the `container` spec are identified with the `test-network-function.com/container=target`
+label. Any value is permitted but `target` is used here for consistency with the `generic` spec.
```
ryandgoulding, line 288: We don't need to address this here, but the cnf-features-deploy test suite targets `worker-cnf` label.  We may want to do something similar, but there may also be value in having a separate label.

ryandgoulding (approved):

PR #206, opened 2021-05-18
CTONET-530 Invoke cnf feature deploy test suite

jun-rh:


ryandgoulding:
This PR requires a git rebase.

ryandgoulding:
We prefer a "rebase-if-necessary" strategy over merge commits.

test-network-function/suite_test.go
```diff
@@ -82,6 +84,18 @@ func createClaimRoot() *claim.Root {
 	}
 }
 
+func loadJunitXMLIntoMap(result map[string]interface{}, filepath, key string) {
+	var err error
+	if key == "" {
+		var extension = path.Ext(filepath)
```
jun-rh, line 90: Accidentally replaced filepath.Ext with path.ext. No real impact when running on linux. Will fix it anyways in the next commit.

run-cnf-suites.sh
```diff
@@ -53,5 +53,18 @@ done
...
+if [ "$VERIFY_CNF_FEATURES" == "true" ] && [ "$TNF_MINIKUBE_ONLY" != "true" ]; then
+
+	export TNF_IMAGE_NAME=cnf-tests
+	export TNF_IMAGE_TAG=latest
+	export TNF_OFFICIAL_ORG=quay.io/openshift-kni/
+	export TNF_OFFICIAL_IMAGE="${TNF_OFFICIAL_ORG}${TNF_IMAGE_NAME}:${TNF_IMAGE_TAG}"
+	export TNF_CMD="/usr/bin/test-run.sh"
+	export OUTPUT_ARG="--junit"
+
+	./run-container.sh -d 127.0.0.53 -n host -o $OUTPUT_LOC -ginkgo.v -ginkgo.skip="performance|sriov|ptp|sctp|xt_u32|dpdk|ovn"
```
jun-rh, line 65: DNS issue resolved after upgrading my docker engine to the latest version. Will remove the -d argument in this line in the next commit.

run-cnf-suites.sh
```diff
@@ -53,5 +53,18 @@ done
 
 FOCUS=${FOCUS%?}  # strip the trailing "|" from the concatenation
 
+if [ "$VERIFY_CNF_FEATURES" == "true" ] && [ "$TNF_MINIKUBE_ONLY" != "true" ]; then
```
crwr45, line 56: Will this work if running within our containerised version? I don't believe we install `docker` in the container (if that works) so I'd be surprised if it does.

jun-rh: It will work if the docker socket is mounted. The first container will be able to start a container on the host. It's tricky with podman though. It's not an officially supported case and it would require --privilege.

jun-rh: Ha, almost lost what I had in mind. I actually had a plan, just didn't recall when reading your comment ... 
Instead of exposing docker socket or running privileged podman, we should invoke cnf feature deploy container in the run-tnf-container script as well. Subsequently, the tnf container itself would always be invoked with "VERIFY_CNF_FEATURES" set to "true". In the future, the CNF specific tests in its own repo/container can be included the same way. The only reason why I didn't do it yet is because the github actions still use minikube anyways. When we have other use cases for running tnf container or the pipeline gets a real ocp lab, we would need this change for sure.

crwr45 (approved):
Remove the need for the merge commit, otherwise this is fine

run-container.sh
```diff
@@ -1,37 +1,12 @@
...
-TNF_OFFICIAL_IMAGE="${TNF_OFFICIAL_ORG}${TNF_IMAGE_NAME}:${TNF_IMAGE_TAG}"
-
 CONTAINER_TNF_DIR=/usr/tnf
 CONTAINER_TNF_KUBECONFIG_FILE_BASE_PATH="$CONTAINER_TNF_DIR/kubeconfig/config"
 CONTAINER_DEFAULT_NETWORK_MODE=bridge
 
 usage() {
 	read -d '' usage_prompt <<- EOF
-	Usage: $0 -t TNFCONFIG -o OUTPUT_LOC [-i IMAGE] [-k KUBECONFIG] [-n NETWORK_MODE] SUITE [... SUITE]
+	Usage: $0 -t TNFCONFIG -o OUTPUT_LOC [-i IMAGE] [-k KUBECONFIG] [-n NETWORK_MODE] [-d DNS_RESOLVER_ADDRESS] SUITE [... SUITE]
```
mkochanowski, line 9: Looking at the `run-cnf-suites.sh` file, the `usage()` helper function should probably be moved from `run-container.sh` to the newly created `run-tnf-container.sh` file.

The usage prompt uses `$0` to display the name of the script (`run-container.sh`).
This means that the usage prompt displays incorrect usage info and examples, e.q.:
```
Usage: ./run-container.sh -t TNFCONFIG -o OUTPUT_LOC [-i IMAGE] [-k KUBECONFIG] [-n NETWORK_MODE] [-d DNS_RESOLVER_ADDRESS] SUITE [... SUITE]

[...]

Examples
  ./run-container.sh -t ~/tnf/config -o ~/tnf/output diagnostic generic

  Because -k is omitted, run-container.sh will first try to autodiscover local kubeconfig files.
  If it succeeds, the diagnostic and generic tests will be run using the autodiscovered configuration.
  The test results will be saved to the '~/tnf/output' directory on the host.
```
Examples provided in the usage message no longer work, as one has to call `./run-tnf-container.sh` instead.

run-container.sh
```diff
@@ -45,6 +20,7 @@ usage() {
 	      The -k option takes precedence, overwriting the results of local kubeconfig autodiscovery.
 	      See the 'Kubeconfig lookup order' section below for more details.
 	  -n: set the network mode of the container.
+	  -d: set the DNS resolver address for the test containers started by docker, may be required with certain docker version if the kubeconfig contains host names
```
mkochanowski, line 23: Not sure if we should break the lines ourselves, or depend on the user's terminal emulator to do it for us. So far, we've been (or rather: I've been) breaking the line at some artificially set point, around the 100 character limit.

We could either:
- break the line, like this:
```suggestion
	  -d: set the DNS resolver address for the test containers started by docker, may be required with
	      certain docker version if the kubeconfig contains host names
```
- or, remove newlines from all lines in the usage prompt broken around the character limit, and just depend on word-wrapping.

mkochanowski:
I've left some additional comments. For some reason they are treated as separate comments, not a part of a single review. Weird.

I strongly believe we should move `./run-container.sh` to some other directory, as it's no longer a script that should be exposed to an end user by being in the root directory, it's only function now is to be a shared and invoked by other scripts, not by the user directly.

We can also rename the script to something else.

Why should we even do that? Well, this PR is introducing a breaking change. Running `./run-container.sh` no longer works with the syntax we've been using so far.
The user will spend less time on debugging if he gets an `"./run-container.sh does not exist"` error, rather than having to debug the use of his arguments for `./run-container.sh`, when in reality, he should be using `./run-tnf-container.sh` instead.

jun-rh:
> I've left some additional comments. For some reason they are treated as separate comments, not part of a single review. Weird.
> 
> I strongly believe we should move `run-container.sh` to some other directory, as it should no longer be exposed to an end user by being in the root directory, it's only function now is to be shared and invoked by other scripts, and not by the user directly.
> 
> We can also rename the script to something else.
> 
> **Why should we even do that?**
> Well, this PR is introducing a breaking change. Running `run-container.sh` no longer works with the syntax we've been using so far.
> 
> The user will spend less time on debugging if he gets an `"run-container.sh does not exist"` error, rather than having to debug the use of his arguments for `run-container.sh`, when in reality, he should be using `run-tnf-container.sh` instead.
> I myself have spent some time trying to understand whether I can still use `run-container.sh`, or not.

Hmm, ok. I thought updating readme and the github actions should be good enough. What name do you have in mind for the utility script? Also, I can change `run-tnf-container` back to `run-container.sh` if it means that much. I debated between the two approaches and picked this one just to make the diff easier to understand. And I didn't like the name `run-container` in the first place. To me, it's a bit too generic and it's kind of perfect for the utility script ...

run-container.sh
jun-rh: Will break the line

run-container.sh
jun-rh: Will move the usage function

mkochanowski:
> Hmm, ok. I thought updating readme and the github actions should be good enough. What name/location do you have in mind for the utility script? Also, I can change `run-tnf-container` back to `run-container.sh` if it means that much. I debated between the two approaches and picked this one just to make the diff easier to understand. And I didn't like the name `run-container` in the first place. To me, it's a bit too generic and it's kind of perfect for the utility script ...

I think the new name `run-tnf-container.sh` is great, it's definitely an improvement compared to `run-container.sh`. 
Thanks for updating the README and especially the GitHub Actions, I appreciate that!

I agree with everything you've said, `run-container` sound like a name for a generic utility script.

With my proposal, I just wanted to make sure we won't have to add an additional question like "which version of run-container.sh are you using?" to our debugging checklist, because the output of the new version of `run-container` looks almost identical to the old one, the only difference being that the new one no longer works for obvious reasons.

---

_If we wish to proceed with the proposed change of moving the generic script to a different location:_

Maybe we could move `run-container.sh` to a directory called `utils/`?
To avoid breaking the diff, moving the shell script can be done in a separate commit.
We already have `fs-diff-test-utils` in the root directory, later we could unify that to achieve a directory structure like this:
```
utils/
	containerized-offering/
		run-container.sh
	fs-diff-test/
		diff-fs.sh
		Dockerfile.podman-image
		privilaged-pod.yaml
```
where `containerized-offering` could be named something else, to better describe this "bucket" of utilities.

What do you think @ryandgoulding, @crwr45? 
We can push theses changes back to be considered and handled in a separate story.

jun-rh:
> > Hmm, ok. I thought updating readme and the github actions should be good enough. What name/location do you have in mind for the utility script? Also, I can change `run-tnf-container` back to `run-container.sh` if it means that much. I debated between the two approaches and picked this one just to make the diff easier to understand. And I didn't like the name `run-container` in the first place. To me, it's a bit too generic and it's kind of perfect for the utility script ...
> 
> I think the new name `run-tnf-container.sh` is great, it's definitely an improvement compared to `run-container.sh`.
> Thanks for updating the README and especially the Actions, I appreciate that!
> 
> I agree with everything you've said, `run-container` sound like a name for a generic utility script.
> 
> With my proposal, I just wanted to make sure we won't have to add an additional question like "which version of run-container.sh are you using?" to our debugging checklist, because the output of the new version of `run-container` looks almost identical to the old one, the only difference being that the new one no longer works for obvious reasons.
> 
> _If we wish to proceed with the proposed change of moving the generic script to a different location:_
> 
> Maybe we could move `run-container.sh` to a directory called `utils/`?
> To avoid breaking the diff, moving the shell script can be done in a separate commit.
> We already have `fs-diff-test-utils` in the root directory, later we could unify that to achieve a directory structure like this:
> 
> ```
> utils/
> 	containerized-offering/
> 		run-container.sh
> 	fs-diff-test/
> 		diff-fs.sh
> 		Dockerfile.podman-image
> 		privilaged-pod.yaml
> ```
> 
> where `containerized-offering` could be named something else, to better describe this "bucket" of utilities.
> 
> What do you think @ryandgoulding, @crwr45?
> We can push back these changes for consideration in a separate story.

Thanks. I like your plan too. We can discuss the actual directory names when we address that new story.

ryandgoulding:
@jun-rh can you please rebase this so we can merge?

jun-rh:
> @jun-rh can you please rebase this so we can merge?

Rebased and added new changes (commit 9666d11) addressing review comments.

ryandgoulding:
Thank you.  I had to manually allow the workflows, since you are a first time contributor, and GitHub requires manual approval for this case.  Assuming that passes, I will merge, and you should not run into these workflow limitations in the future.  Thank you for the contribution!

jun-rh:
> Thank you. I had to manually allow the workflows, since you are a first time contributor, and GitHub requires manual approval for this case. Assuming that passes, I will merge, and you should not run into these workflow limitations in the future. Thank you for the contribution!

Just realized that I was using my old github account when committing/pushing. I did an amend commit and corrected the last commit. Let me know if I need to re-commit both.

jun-rh:
@ryandgoulding, just pushed a fix for the bug I found after the rebase. It seems that workflow approval is needed. It's probably because I am using the right account now. The approval you did might be for my old account.

PR #209, opened 2021-05-20
Fix broken pipeline

ryandgoulding:
This is considered an emergent fix in order to unblock the GitHub pipeline.
As such, review comments will be addressed in a follow-up rather than prior to
merging.

ubi8/ubi maintainers recently released a new version of the base container
image, which contains an updated yum manifest.  `golangci-1.14.12` is no
longer a valid package in `latest`.

This change does the following:
0) Checks to ensure the docker image is built on top of `x86_64` architecture.
   If not, the build exits with an appropriate error message.
1) Removes the `golang-1.14.12` package from the yum install list.  Since this
   package also brought in `gcc` and `krb5-devel`, which are compile-time
   dependencies for `oc`, `gcc` and `krb5-devel` were added to the yum install
   list.  `oc` would not compile otherwise.  Additionally, `wget` as added to
   the yum install list in order to GET the golang binary.
2) Installs the golang-1.14.12 binary from the golang downloads site.
3) Sets `$PATH` to include /usr/local/go/bin and `$GOHOME`/bin.  The latter of
   which was automatically set as a part of the yum installation process, and
   is explicitly needed when installing the binary manually.  Otherwise,
   `mockgen` does not propagate to the `$PATH` properly.
4) Cleans up additional added compile time dependencies from step 1 above,
   keeping runtime dependencies (just `jq`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM registry.access.redhat.com/ubi8/ubi:latest AS build
 
-# golang 1.14.12 as it's the only version available on yum at time of writing.
-ENV GOLANG_VERSION=1.14.12
 ENV GOLANGCI_VERSION=v1.32.2
 
 ENV TNF_DIR=/usr/tnf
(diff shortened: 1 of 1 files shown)


ryandgoulding:
Folks,

As the lead maintainer for this repository, I am making the executive decision to self-merge this code, as no one is around for a review.  I am fully dedicated to incorporating any feedback after the fact.  This is being merged as an emergent fix to unblock our pipeline.  I do not enjoy self-merging patches, and will not make a habit of it when possible.

-Ryan

crwr45 (commented):
lgtm

PR #210, opened 2021-05-21
Autodiscover for operator config

crwr45:


pkg/config/autodiscover/csv_info.go
```diff
@@ -0,0 +1,68 @@
+package autodiscover
```
ryandgoulding, line 1: Please add a copyright header.

pkg/config/autodiscover/autodiscover_operator.go
```diff
@@ -0,0 +1,69 @@
+package autodiscover
```
ryandgoulding, line 1: Same, please add the license to the header.

pkg/config/autodiscover/csv_info.go
crwr45: Seems like I have a blind spot! Copying the wrong one last time, omitting it completely this time. Thanks for spotting it!

ryandgoulding (approved):

PR #212, opened 2021-05-23
Change container tooling and docs to use podman, and not to run as root

shimritproj:
Change container tooling and docs to use podman, and not to run as root


shimritproj:
I have PR update ("776-Done!")

PR #213, opened 2021-05-25
CTONET-880 - make sure MCO does not contradict with sysctl config files

isaacdorfman:


test-network-function/generic/suite.go
```diff
@@ -510,6 +586,21 @@ func testBootParams(context *interactive.Context, podName, podNamespace string,
...
 }
 
+func testSysctlConfigs(context *interactive.Context, podName, podNamespace string) {
+	ginkgo.It(fmt.Sprintf("Testing sysctl config files for the pod's node %s/%s", podNamespace, podName), func() {
+		nodeName := getPodNodeName(context, podName, podNamespace)
+		combinedSysctlSettings := getSysctlConfigArgs(context, nodeName)
+		mcName := getMcName(context, nodeName)
+		mcKernelArgumentsMap := getMcKernelArguments(context, mcName)
+
+		for key, sysctlConfigVal := range combinedSysctlSettings {
```
crwr45, line 596: I don't know enough about how the two different sets of values interact to know if this is correct but currently this is not asserting in all cases.

Currently if a value is:
1. present in both maps and equal -> assertion passes
2. present in both maps but unequal -> assertion fails
3. present only in `sysctlConfigVal` -> triggers no assertion, therefore passes
4. present only in `mcKernelArgumentsMap` -> triggers no assertion, therefore passes

@ryandgoulding do you know if the current behaviour is correct?

pkg/tnf/handlers/readremotefile/readremotefile.go
```diff
@@ -0,0 +1,97 @@
...
+
+// ReadRemoteFile holds information regarding remote file contents at a specified node and path.
+type ReadRemoteFile struct {
+	remoteFileContents string // Output variable that stores the remote file contents
+	result             int
+	timeout            time.Duration
+	args               []string
+}
+
+// NewReadRemoteFile creates a ReadBootConfig tnf.Test.
```
crwr45, line 39: update comment

crwr45 (changes requested):

test-network-function/generic/suite.go
isaacdorfman: You understand it correctly.
This is the way it supposed to behave.

test-network-function/generic/suite.go
```diff
@@ -420,13 +427,17 @@ func getMcName(context *interactive.Context, nodeName string) string {
 	return mcNameTester.GetMcName()
 }
 
+func runAndValidateTest(test *tnf.Test) {
```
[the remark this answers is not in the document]
isaacdorfman: Done

test-network-function/generic/suite.go
```diff
@@ -491,6 +502,71 @@ func getGrubKernelArgs(context *interactive.Context, nodeName string) map[string
...
+	for _, line := range splitConfig {
+		if line == "" {
+			continue
+		}
+
+		if strings.HasPrefix(line, "#") || strings.HasPrefix(line, ";") {
+			continue
+		}
+
+		keyValRegexp := regexp.MustCompile("(\" \")*\\S*(\" \")*\\S*(\" \")*")
```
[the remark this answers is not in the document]
isaacdorfman: I changed the regex.

ryandgoulding:
This fails smoke tests, and will not work on minikube.  Please add logic to skip for minikube environment.

PR #216, opened 2021-05-25
Ctonet 883

wying3:
Remove use of deprecated Ginkgo APIs and fix the JunitPath


crwr45:
At a glance, your changes in the final commit look good, but this PR should be a single commit containing only those changes. Please **`rebase`** without rewriting all the commits so that there are no extra changes from `main`. The amount of changes (2000 changes across 50 files) makes this un-reviewable in its current state.

wying3:
fixed now with 1 commit. thanks Charlie's comments

ryandgoulding (approved):

PR #222, opened 2021-05-26
Create descriptions/remediations for each JUnit test case

ryandgoulding:
Augment the catalog generation to include the test information.
Additionally, regenerate CATALOG.md to include all the new great
tests that were added in CNF Certification 2.0.

Additionally, implement results recording for non-generic test
suites.

### DIFF
--- a/cmd/catalog/cmd/generate.go
+++ b/cmd/catalog/cmd/generate.go
@@ -19,8 +19,12 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path"
 	"strings"
+
+	"github.com/test-network-function/test-network-function/test-network-function/identifiers"
 
 	"github.com/spf13/cobra"
 	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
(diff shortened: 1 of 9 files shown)


cmd/catalog/cmd/generate.go
```diff
@@ -85,9 +86,33 @@ func cmdJoin(elems []string, sep string) string {
 
 // runGenerateMarkdownCmd generates a markdown test catalog.
 func runGenerateMarkdownCmd(_ *cobra.Command, _ []string) error {
-	fmt.Println("# `tnf.Test` Catalog")
+	fmt.Println("# test-network-function test case catalog")
```
crwr45, line 89: Is there a reason this section can't be replaced with a multi-line string?

crwr45 (changes requested):

cmd/catalog/cmd/generate.go
ryandgoulding: I made some heavy-handed changes to this so that static text is loaded from file rather than generated directly via Go.  The dynamic text is still generated in Go, and I would prefer to keep the multiple invocations of `fmt.Println(...)` and `fmt.Sprintf(...)` as they keep each line concise and easy to review.  The only other option I can see is to use a `Go text/template`, but that is fairly heavy-handed for something like this.

crwr45 (approved):

PR #231, opened 2021-06-02
CTONET-999: Resolves issues with generic test suite

ryandgoulding:
The generic test suite was failing the Pod ownership test.  This change
resolves a few issues with the test including:

1) The ReelFirst() regular expression was made to be more specific.  Namely,
   the test now expects the proper "OWNERKIND" header, and parses output after
   ignoring the header line.
2) The "-o custom-columns"... line was modified to properly escape brackets and
   the asterisk character.  In many versions of `oc`, these characters must be
   either quoted or escaped.
3) Existing positive unit tests seeded test data from the negative test cases.
   This change switches to accurate positive test data mock output.
   Additionally, the requirement of an exact match is dropped due to the fact
   that trailing newlines may occur with the modified regular expression from
   step 1.

I have tested this change locally several times to ensure that the results are
not sporadically failing.

### DIFF
--- a/pkg/tnf/handlers/deploymentsnodes/deploymentsnodes_test.go
+++ b/pkg/tnf/handlers/deploymentsnodes/deploymentsnodes_test.go
@@ -71,7 +71,7 @@ func Test_ReelMatchSuccess(t *testing.T) {
 }
 
 // Just ensure there are no panics.
-func Test_ReelEof(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ryandgoulding:
This resolves an issue that is causing GitHub Workflows to fail consistently for every new PR, and blocks the code pipeline.  As such, I am going to merge this to unblock the pipeline.  I am happy to review follow-up comments and address any desired changes.

PR #242, opened 2021-06-07
CTONET-530 Fix the issue from the podman change and also invoke the t…

jun-rh:
…est suite in run-tnf-container.sh

### DIFF
--- a/run-cnf-suites.sh
+++ b/run-cnf-suites.sh
@@ -53,29 +53,12 @@ done
 
 FOCUS=${FOCUS%?}  # strip the trailing "|" from the concatenation
 
-if [ "$VERIFY_CNF_FEATURES" == "true" ] && [ "$TNF_MINIKUBE_ONLY" != "true" ]; then
-
-	export TNF_IMAGE_NAME=cnf-tests
-	export TNF_IMAGE_TAG=latest
-	export TNF_OFFICIAL_ORG=quay.io/openshift-kni/
-	export TNF_OFFICIAL_IMAGE="${TNF_OFFICIAL_ORG}${TNF_IMAGE_NAME}:${TNF_IMAGE_TAG}"
-	export TNF_CMD="/usr/bin/test-run.sh"
-	export OUTPUT_ARG="--junit"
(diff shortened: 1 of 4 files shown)


jun-rh:
Forgot to update the image docker file for the script location change. Will update this.

jun-rh:
@ryandgoulding @crwr45

run-cnf-suites.sh
```diff
@@ -53,29 +53,11 @@ done
...
-	fi
-	
-	# For older verions of docker, dns server may need to be set explicitly, e.g.
-	#
-	# export DNS_ARG=172.0.0.53
-	./run-container.sh -ginkgo.v -ginkgo.skip="performance|sriov|ptp|sctp|xt_u32|dpdk|ovn"
-else
-	# removing report if not running, so the final claim won't include stale reports
-	rm -f ${OUTPUT_LOC}/validation_junit.xml
+if grep -q init\.scope /proc/1/cgroup; then
```
crwr45, line 56: This doesn't handle running on a Mac well:
```
cwheeler@MacBook-Pro test-network-function % TNF_ENABLE_CONFIG_AUTODISCOVER=true ./run-cnf-suites.sh generic
grep: /proc/1/cgroup: No such file or directory
```

crwr45 (changes requested):
Looks good, one issue running on mac

run-cnf-suites.sh
jun-rh: hmm, i can't test this as i don't have a mac. do you have this file when running in container then? i could do the opposite check. On linux, this file would contain something like "docker" or "user.slice".

jun-rh: @crwr45 @ryandgoulding Can one of you check the existence and the content of /proc/1/cgroup when running docker and podman on your MAC? Thanks.

jun-rh: Ok. I will use the absence of the file as another indicator of not running in container.

ryandgoulding (approved):

PR #245, opened 2021-06-08
Automatic terminal prompt emulation

ryandgoulding:
CTONET-1013 and CTONET-1011 both share a common problem;  it appears that
results are being truncated.  This is true, but it is supposed to happen.  The
way `expect` works is to return the first match of a given regular expression.
Even if greedy quantifiers such as `*` are utilized, google/goexpect is based
on copying an output buffer every so often (the interval is identified by
chkDuration duration.Duration in goexepct).

As such, consider the following regular expression:

`(?s).+`

This could match something as simple as:

`BeginningOfTheOutput...`

When the user fully intended to capture:

```
BeginningOfTheOutput...
EndOfTheOutput
```

The problem compounds for situations in which the command output is arbitrary
in length, and could very well not end with a sentinel at all.  In typical
expect programs, users can just expect the terminal prompt.  In this case
though, google/goexpect delegates consumption of the prompt to the underlying
PTY, so we cannot simply expect the terminal prompt.  As such, we must get a
bit more creative.

This PR introduces the ability to postfix any command with a sentinel,
indicating that the command output has completed successfully.  This allows
us to fully drain the google/goexpect PTY buffers, and greedily capture as much
output as possible.

The mechanism is on by default, but can be turned off per-Test or per-Reel.
The idea is to emulate the resulting terminal prompt which visually lets a user
know that the previous command has completed.  Think of something similar to
the following:

```
[redhat@bastion ~]$ ls /tmp
systemd-private-5a3e10cedce0468fa62fa3e366810b8c-chronyd.service-T66lA8
systemd-private-5a3e10cedce0468fa62fa3e366810b8c-httpd.service-TWeooL
systemd-private-5a3e10cedce0468fa62fa3e366810b8c-systemd-resolved.service-tGDXYz
tmux-1007
[redhat@bastion ~]$
```

We only know the command is done when we see the "[redhat@bastion ~]" prompt.

Unit tests have been augmented.  Cases were modified for both situations in
various places to ensure both strategies work correctly.

### DIFF
--- a/pkg/tnf/reel/reel.go
+++ b/pkg/tnf/reel/reel.go
@@ -17,6 +17,7 @@
 package reel
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


crwr45 (approved):
Are the changes to the configuration intended?

ryandgoulding:
> Are the changes to the configuration intended?

No, good catch.  I have reset the `tnf_config.yaml` changes and force-pushed a new copy.

PR #247, opened 2021-06-09
Add in functionality to uncordon nodes

ryandgoulding:
Uncordon nodes that are drained so that they are not left idle.
This code is unit tested.  The test is implemented using a JSON
template, and a new API was added to utilize a map to fill in the
template values.

A few identifiers were fixed as a part of this commit.

Finally, TNF_NON_INTRUSIVE_ONLY was added to only run Non-intrusive
test specs.


crwr45 (approved):

ryandgoulding:
Just a rebase...

Issue #233, opened 2021-06-02
Incorrect message when CertifiedOperatorInfo is missing


betoredhat:
Hello,

When CertifiedOperatorInfo is missing from the config file, it complains about missing CertifiedOperatorRequestInfo. 

```
should be created without error [It]                                                                                                                                        [149/1855]
   test-network-function/test-network-function/operator/suite.go:62                                                                                             
                                                                                                                                                                                          
    Expected                                                                                                                                                                              
        <[]configsections.CertifiedOperatorRequestInfo | len:0, cap:0>: nil                                                                                                               
    not to be nil  
```
Looks like something related to this bue unfortunately my GO fu is not strong. 
```
 grep CertifiedOperatorRequestInfo pkg/config/config.go 
        CertifiedOperatorInfo []configsections.CertifiedOperatorRequestInfo `yaml:"certifiedoperatorinfo,omitempty" json:"certifiedoperatorinfo,omitempty"`
```

Regards

ryandgoulding:
Any chance you can provide the config file @betoredhat?  If you wish to skip those tests, I suggest deleting the entire "certifiedoperatorinfo" section.

betoredhat:
sure, this is the file:
```
operators:
  - name: testpmd-operator.v0.2.4
    namespace: example-cnf
    status: Succeeded
    subscriptionName: testpmd-operator-subscription
    autogenerate: false
    tests:
      - OPERATOR_STATUS
certifiedoperatorinfo:
  - name: testpmd-operator
    organization: xyz
cnfs:
  - name: testpmd-operator-controller-manager-54cd68897f-sllvl
    namespace: example-cnf
    status: Running
    tests:
      - PRIVILEGED_POD
      - PRIVILEGED_ROLE
    certifiedcontainerrequestinfo:
  - name: testpmd-app-7875b84547-bw8qk
    namespace: example-cnf
    status: Running
    tests:
      - PRIVILEGED_POD
      - PRIVILEGED_ROLE
    certifiedcontainerrequestinfo:
  - name: testpmd-app-7875b84547-mkmfb
    namespace: example-cnf
    status: Running
    tests:
      - PRIVILEGED_POD
      - PRIVILEGED_ROLE
    certifiedcontainerrequestinfo:
```
if I do not set the following section it will complain saying that CertifiedOperatorRequestInfo is missing. But it is actually "certifiedoperatorinfo" that looks like mandatory.
```
certifiedoperatorinfo:
[...]

PR #252, opened 2021-06-14
Query for operator/container certification only when present

ryandgoulding:
Only query for container/operator certification when they are present.  This
ultimately makes these sections of config optional.


ryandgoulding:
@jun-rh please take a look at this as well.  Thank you.

jun-rh (approved):
The only change is the new if check, right?

ryandgoulding:
yes

ryandgoulding:
Resolves #233

PR #253, opened 2021-06-15
Resolves issues with checking operator subscription

ryandgoulding:
Hitherto, the operator subscription is tested against a non-templated
test case input file.  The existing framework does not have the innate
ability to render the file contents based on a template, making it an
unsuitable candidate for a test which requires specific, templated output.
Since test cases ultimately are fed into the tool using a Go string,
templating this code seemed like the wrong solution.

This change rips the "CHECK_SUBSCRIPTION" test out of the existing operator
testing framework.  That framework is difficult to maneuver, and this change
aligns with utilizing the JSON test framework instead.  Why was this approach
chosen?  Here are some of the reasons:

1) Standardize on JSON Test Template framework.
2) Disaggregate the reference to the "operator-is-installed-via-olm" from the
   existing "operator-best-practices" test case.  This provides further
   granularity so that this test case can pass/fail independently of the other
   operator best practices.
3) Unit testing is crafted towards the specific use case, instead of testing
   pseudo-template rendering.  In this case, we can use actual input/output
   that the Handler would encounter.
4) The code is easier to read.

Finally, CATALOG.md is regenerated to reflect the changes.  This involved
eliminating the "pod-is-recreated" test case from the catalog, which was
removed in an earlier PR.

### DIFF
--- a/test-network-function/operator/suite.go
+++ b/test-network-function/operator/suite.go
@@ -18,8 +18,11 @@ package operator
 
 import (
 	"fmt"
+	"path"
 	"strings"
 	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/generic"
 
 	"github.com/test-network-function/test-network-function/test-network-function/identifiers"
 	"github.com/test-network-function/test-network-function/test-network-function/results"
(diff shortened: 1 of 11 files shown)


ryandgoulding:
@jun-rh please take a look.

test-network-function/operator/suite.go
```diff
@@ -71,9 +89,47 @@ var _ = ginkgo.Describe(testSpecName, func() {
...
 
+// testOperatorsAreInstalledViaOLM ensures all configured operators have a proper OLM subscription.
+func testOperatorsAreInstalledViaOLM() {
+	_, operatorsInTest := getConfig()
+	for _, operatorInTest := range operatorsInTest {
+		ginkgo.Context("an operator is installed", func() {
+			ginkgo.When("subscriptions are polled", func() {
+				ginkgo.It(fmt.Sprintf("%s in namespace %s Should have a valid subscription", operatorInTest.SubscriptionName, operatorInTest.Namespace), func() {
+					defer results.RecordResult(identifiers.TestOperatorIsInstalledViaOLMIdentifier)
+					testOperatorIsInstalledViaOLM(operatorInTest.SubscriptionName, operatorInTest.Namespace)
```
jun-rh, line 104: For a second, I thought we were doing a recursive call

jun-rh (approved):
We can discuss this in the KT session tomorrow. I am not sure whether the unit test provides value for json based generic test cases.

ryandgoulding:
Resolves #253

ryandgoulding:
The value is in using realistic mock output.  In fact, copying and pasting output from a command is best.  I could envision automatically generating this data as part of a "wizard" at some point.  That said, it gives us some level of confidence that the `tnf.Test` works as expected.

PR #259, opened 2021-06-23
PR for CNFCERT-9

wying3:


jun-rh:
@wying3 please fix the linter errors. Also double check the change list. Not sure what those empty files are. The gradetool binary was included by mistake?

jun-rh (commented):
A general comment, have you considered doing this test with json only? it might actually be easier, less change required.

pkg/tnf/handlers/podselector/podselector.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"time"
+
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	nsRegex = "<none>\\s*<none>"
```
jun-rh, line 28: the point is to make sure there IS a pod anti-affinity rule. additionally, it probably makes sense to check that there is no affinity rule.

jun-rh (changes requested):

test-network-function/identifiers/identifiers.go
```diff
@@ -306,6 +313,18 @@ to why nodeSelector and/or nodeAffinity is utilized by a CNF.`,
 instantiation on any underlying Node.`),
 	},
 
+	TestPodSelectorAndAffinityBestPractices: {
+		Identifier: TestPodSelectorAndAffinityBestPractices,
+		Type:       informativeResult,
+		Remediation: `In most cases, Pod's should not specify their host Nodes through nodeSelector or nodeAffinity.  However, there are
```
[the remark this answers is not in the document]
wying3: redo the code using json based approach

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
```diff
@@ -0,0 +1,29 @@
+{
+  "identifier": {
+    "url": "http://test-network-function.com/tests/podantiaffinity",
+    "version": "v1.0.0"
+  },
+  "description": "This test checks cnf pod antiaffinity rule in cnf deployment.",
+  "testResult": 0,
+  "testTimeout": 10000000000,
+  "reelFirstStep": {
+    "execute": "oc get deployment {{.DEPLOYMENT_NAME}} -n {{.DEPLOYMENT_NAMESPACE}} -o json | jq -r ' if .spec.replicas > 1 then if .spec.template.spec.affinity.podAntiAffinity != null then \"OK\" else \"MISSING\" end else \"OK\" end '\n",
```
jun-rh, line 10: this is probably not recommended. It would be cleaner to have two handlers

pkg/tnf/identifier/identifiers.go
```diff
@@ -51,6 +51,7 @@ const (
 	readRemoteFileIdentifierURL           = "http://test-network-function.com/tests/readRemoteFile"
 	uncordonNodeIdentifierURL             = "http://test-network-function.com/tests/node/uncordon"
 	checkSubscriptionIdentifierURL        = "http://test-network-function.com/tests/operator/check-subscription"
+	podantiaffinityIdentifierURL          = "http://test-network-function.com/tests/podantiaffinity"
```
jun-rh, line 54: As I mentioned in standup, it's better to keep the test case name more high level and customer oriented, e.g. testPodHighAvailability. The handlers can be named with specific checks, like testPodAntiaffinity, testReplicaCount

test-network-function/identifiers/identifiers.go
```diff
@@ -299,13 +306,24 @@ the same hacks.'`),
...
 		Description: formDescription(TestPodNodeSelectorAndAffinityBestPractices,
 			`ensures that CNF Pods do not specify nodeSelector or nodeAffinity.  In most cases, Pods should allow for
 instantiation on any underlying Node.`),
 	},
 
+	TestPodAntiAffinityBestPractices: {
+		Identifier: TestPodAntiAffinityBestPractices,
+		Type:       informativeResult,
+		Remediation: `In most HA cases, Pod podAntiAffinity rule should be specified when replica value is set during pod instantiation. 
+		However, there are cases in which CNFs require specialized hardware specific to a particular class of Node.  As such,this test is purely informative
```
jun-rh, line 320: This sentence does not apply here. It's for node selector/affinity. Maybe get ride of the rest too. This is a pretty firm rule for HA, unlike the node selector check.

jun-rh (changes requested):

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
HamadiSE: @jun-rh  probably something like this ?
```
oc get -n sqlnamespace deployment/sql -o go-template='{{ .spec.replicas }}{{" "}}{{len  .spec.template.spec.affinity.podAntiAffinity }}{{"\n"}}'
```

pkg/tnf/identifier/identifiers.go
wying3: you mean URL "podantiaffinity" = >  testPodHighAvailability?

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
wying3: @HamadiSE the go-template syntex is not working when podAntiAffinity  is not present in json.
@jun-rh , not fully get the need to have two handlers when one query and filter can process the the context.

test-network-function/identifiers/identifiers.go
wying3: how about this:  
"In high availability case, podAntiAffinity rule should be specified for pod instantiation if pod replica value is set more than 2 .

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
jun-rh: @wying3 In general, this kind of code embedded in a command filter is not very reusable or extensible. I had a hard time trying to tell whether the return values make sense from this one line nested if block. Is the last ok returned when replica count is 1? If that's the case, it would be wrong. We want to fail the test if the count is not greater than 1. Think about maintainability, would you continue this approach when we need to add a third or fourth check?

jun-rh: BTW, in theory, we need to check not just the presence of the antiaffinity rule, but also the content of the rule. There has to be at least one rule that is against the pod's own label with RequiredDuringScheduling and Node topology. Can this be done in a jq expression in a clean way? Maybe even the json based generic test becomes too limited for that kind of check.

pkg/tnf/identifier/identifiers.go
jun-rh: yes

test-network-function/identifiers/identifiers.go
jun-rh: i would say pod scheduling instead of pod instantiation.

jun-rh:
New test case failing in smoke test. Update the partner deployment yaml in the other repo.

pkg/tnf/handlers/podantiaffinity/podantiaffinity_test.go
```diff
@@ -0,0 +1,141 @@
...
+	_, handlers, jsonParseResult, err := createTest()
+
+	assert.Nil(t, err)
+	assert.True(t, jsonParseResult.Valid())
+	assert.NotNil(t, handlers)
+
+	assert.Equal(t, 1, len(handlers))
+	handler := handlers[0]
+	step := handler.ReelFirst()
+	expectedReelFirstExecute := fmt.Sprintf("oc get deployment %s -n %s -o json | jq -r ' if .spec.replicas > 1 then if .spec.template.spec.affinity.podAntiAffinity != null then \"OK\" else \"PODAAFAIL\" end else \"REPLICAFAIL\" end '\n",
```
jun-rh, line 86: wrap the line at a reasonable width

pkg/tnf/identifier/identifiers.go
jun-rh: test name/url not updated

jun-rh (changes requested):

test-network-function/generic/suite.go
```diff
@@ -885,6 +896,51 @@ func uncordonNode(node string) {
 	gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
 }
 
+// Pod antiaffinity test for all deployments
+func testPodAntiAffinity(podNamespace string) {
+	var deployments dp.DeploymentMap
+	ginkgo.When("CNF is designed in high availability mode ", func() {
+		ginkgo.It("Should have pod replica number set great than 1 and anti-affinity rule been defined in deployment", func() {
```
jun-rh, line 903: *greater*
*pod anti-affinity*

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
```diff
@@ -0,0 +1,34 @@
...
+      "(?m)OK",
+      "(?m)PODAAFAIL",
+      "(?m)REPLICAFAIL"
+    ],
+    "timeout": 10000000000
+  },
+  "resultContexts": [
+    {
+      "pattern": "(?m)OK",
+      "defaultResult": 1
```
jun-rh, line 21: I see failure logs like this right now:

  Expected
        <int>: 2
    to equal
        <int>: 1

It doesn't show what result was causing code 2. @HamadiSE, can we enhance the implementation for json based test and make it print the actual pattern matched instead of just the code?

jun-rh (changes requested):

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
HamadiSE: I think so. There's the generic.go that handles this and can be extended.
Also, with the current deployment files, this test will fail because we don't have replicas or anti affinity defined.

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    {
+      "pattern": "(?m)PODAAFAIL",
+      "defaultResult": 2
+    },
+    {
+      "pattern": "(?m)REPLICAFAIL",
+      "defaultResult": 2
```
HamadiSE, line 29: I think it's best practice "invented by me" to add a 
```
{
      "pattern": "(?m)[.*]",
      "defaultResult": 0
    }
```
in the end, this will catch everything else and return an ```error```

test-network-function/generic/suite.go
wying3: thanks for catch this

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
HamadiSE: my comment get posted twice, then when I delete it, it gets deleted from both places -I need to get familiar with this github better-

regarding the comments, we can use the generic.go to be more verbose in case of failure, we can print the expected and the matched pattern

HamadiSE: https://issues.redhat.com/browse/CNFCERT-46 created

jun-rh: Thanks. @wying3 Make the error string/pattern more readable and descriptive. Once the task Salah created is done, it will show up in the log.

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
wying3: "0" will be considered as tnf.ERROR, replica =1 is not error but  failure, right?
i think i can add one more pattern as you list above for  tnf.ERROR

HamadiSE: This is to catch anything unexpected, for example a failure in executing the command...

jun-rh: @HamadiSE Shouldn't this behavior realized in code instead of json?

HamadiSE: @jun-rh That's also a solution. We should just adhere to the principal that the default return code is error.

jun-rh: Yes

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
jun-rh: @wying3 can you change "PODAAFAIL" and "REPLICAFAIL" to something more descriptive?

pkg/tnf/handlers/podantiaffinity/podantiaffinity_test.go
jun-rh: You could break long string:

"line 1" +
"line 2" +
"line 3"

test-network-function/identifiers/identifiers.go
```diff
@@ -314,13 +321,21 @@ the same hacks.'`),
...
 to why nodeSelector and/or nodeAffinity is utilized by a CNF.`,
 		Description: formDescription(TestPodNodeSelectorAndAffinityBestPractices,
 			`ensures that CNF Pods do not specify nodeSelector or nodeAffinity.  In most cases, Pods should allow for
 instantiation on any underlying Node.`),
 	},
 
+	TestPodHighAvailabilityBestPractices: {
+		Identifier:  TestPodHighAvailabilityBestPractices,
+		Type:        informativeResult,
+		Remediation: `In high availability cases, Pod podAntiAffinity rule should be specified for pod scheduling if pod replica value is set more than 1 .`,
```
jun-rh, line 334: and pod replica should be set, not if

jun-rh (changes requested):

jun-rh:
General comment, please stop pushing amended commits. It wipes off older version and make it really hard to review comment resolution.

test-network-function/identifiers/identifiers.go
wying3: updated

HamadiSE (commented):
> General comment, please stop pushing amended commits. It wipes off older version and makes it really hard to review comment resolution.

Also, please update the catalog.md
you can use this command to generate the new catalog.md

`go run  cmd/catalog/main.go generate markdown > CATALOG.md`

jun-rh (approved):

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
```diff
@@ -0,0 +1,37 @@
+{
+  "identifier": {
+    "url": "http://test-network-function.com/tests/testPodHighAvailability",
+    "version": "v1.0.0"
+  },
+  "description": "This test checks cnf pod antiaffinity rule in cnf deployment.",
+  "testResult": 0,
```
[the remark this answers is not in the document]
jun-rh: @HamadiSE Thanks for confirming. That's what I really meant by code, i.e. the generic test handler implementation.

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
jun-rh: @wying3 I guess you missed this comment?

wying3: ok, will make changes

wying3: how about change to ANTIAFFINITYMISSING  and REPLICAEQ1?

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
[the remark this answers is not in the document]
wying3: will remove arbitrary match

pkg/tnf/handlers/podantiaffinity/podantiaffinity_test.go
wying3: changed to 3 short lines

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
wying3: new match pattern: replica count is 1 and Antiaffinity missing

pkg/tnf/handlers/podantiaffinity/podantiaffinity.json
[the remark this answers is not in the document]
wying3: removed arbitrary match

test-network-function/generic/suite.go
```diff
@@ -885,6 +896,53 @@ func uncordonNode(node string) {
...
+	gomega.Expect(handlers).ToNot(gomega.BeNil())
+	gomega.Expect(len(handlers)).To(gomega.Equal(1))
+	gomega.Expect(test).ToNot(gomega.BeNil())
+	tester, err := tnf.NewTest(context.GetExpecter(), *test, handlers, context.GetErrorChannel())
+	gomega.Expect(err).To(gomega.BeNil())
+	gomega.Expect(tester).ToNot(gomega.BeNil())
+
+	testResult, err := tester.Run()
+	if testResult != tnf.SUCCESS {
+		msg := fmt.Sprintf("The POD replica is found %d, and podAntiAffinity rule is not defined, you might want to change it in deployment %s in namespace %s",
```
edcdavid, line 938: and podAntiAffinity rule is not defined, 
->
but podAntiAffinity rule is not defined or replica count is one,

pkg/tnf/handlers/podantiaffinity/podantiaffinity_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// Positive Test
+	step := handler.ReelMatch(expectedPassPattern, "", "OK")
+	assert.Nil(t, step)
+	assert.Equal(t, tnf.SUCCESS, (*tester).Result())
+
+	// Negative Test
+	step = handler.ReelMatch(expectedFailPattern1, "", "Antiaffinity missing")
+	assert.Nil(t, step)
+	assert.Equal(t, tnf.FAILURE, (*tester).Result())
```
edcdavid, line 139: Just a comment, no need to change. Seems like for simple json tests, we could skip unit testing altogether? This testing does not seem very useful as we pass input and output directly

test-network-function/generic/suite.go
edcdavid: Thanks for the clarification Ying, please disregard previous comment. What about the following change:

if testResult != tnf.SUCCESS {
     if replica >1 {
		msg := fmt.Sprintf("The deployment replica count is %d, but a podAntiAffinity rule is not defined, you might want to change it in deployment %s in namespace %s",
replica, deployment, podNamespace)
} else {
                msg := fmt.Sprintf("The deployment replica count is %d. Pod replica should be >2 with an podAntiAffinity rule defined . You might want to change it in deployment %s in namespace %s",
replica, deployment, podNamespace)
}

wying3: looks better, will make change

pkg/tnf/handlers/podantiaffinity/podantiaffinity_test.go
edcdavid: The unit test might still be required to test the json file itself

jun-rh: I had the same question the first time I saw this json based generic test handler implementation. Ryan answered my question in the last training video. I wasn't fully convinced. Need to look into the actual code to understand how to unit test it better.

test-network-function/generic/suite.go
```diff
@@ -935,10 +935,18 @@ func podAntiAffinity(deployment, podNamespace string, replica int) {
 
 	testResult, err := tester.Run()
 	if testResult != tnf.SUCCESS {
-		msg := fmt.Sprintf("The POD replica is found %d, and podAntiAffinity rule is not defined, you might want to change it in deployment %s in namespace %s",
-			replica, deployment, podNamespace)
-		log.Warn(msg)
-		infoWriter(msg)
+		if replica > 1 {
```
jun-rh, line 938: Good change. I was thinking about this too. Again, if we had two separate check functions, this wouldn't feel like additional work.

jun-rh (approved):

edcdavid (approved):
ok for me

pkg/tnf/identifier/identifiers.go
```diff
@@ -530,6 +531,19 @@ var Catalog = map[string]TestCatalogEntry{
...
+		Description: "A generic test used to check pod's replica and podAntiAffinity configuration in high availability mode",
+		Type:        Normative,
+		IntrusionSettings: IntrusionSettings{
+			ModifiesSystem:           false,
+			ModificationIsPersistent: false,
+		},
+		BinaryDependencies: []string{
+			dependencies.OcBinaryName,
+			// dependencies.GrepBinaryName,
+		},
```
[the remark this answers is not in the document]
wying3: updated

PR #261, opened 2021-06-25
CNFCERT-34 Custom label and pod/container re-discover support

jun-rh:


jun-rh:
@ydayagi @isaacdorfman Please take a look.

pkg/config/config.go
```diff
@@ -41,8 +41,18 @@ const (
...
+type Label struct {
+	Namespace string `yaml:"namespace" json:"namespace"`
+	Name      string `yaml:"name" json:"name"`
+	Value     string `yaml:"value" json:"value"`
+}
+
 // File is the top level of the config file. All new config sections must be added here
 type File struct {
+	// Custom Pod labels for discovering containers under test for generic and container suites
+	TargetPodLabels []Label `yaml:"targetPodLabels,omitempty" json:"targetPodLabels,omitempty"`
```
crwr45, line 54: As implemented adding a label here is equivalent to setting pods with that label as a test target in both the `Generic` and `CNFs` sections of the configuration. This is a substantial change (and improvement) from the previous approach of completely separate configurations for test specs and I want to confirm that is the intention.

I strongly suggest that the rest of the configuration is updated to match this paradigm shift sooner rather than later to avoid confusion and fragmentation. Maintaining configuration in the siloed and unified approaches simultaneously for any longer than necessary is likely to cause problems.

crwr45: Add an explanation to the config docs to document the purpose of this new top-level configuration section.

crwr45 (changes requested):
Given the increased use of label namespace(s) throughout this patch I suggest renaming the `labelNamespace' constant to something more distinct from the `labelNS` parameter introduced. This will help significantly with readability. e.g:
```go
tnfNamespace = "test-network-function.com"
```

jun-rh:
@crwr45 Thanks, all very good points.

pkg/config/config.go
jun-rh: > As implemented adding a label here is equivalent to setting pods with that label as a test target in both the `Generic` and `CNFs` sections of the configuration. This is a substantial change (and improvement) from the previous approach of completely separate configurations for test specs and I want to confirm that is the intention.
> 
> I strongly suggest that the rest of the configuration is updated to match this paradigm shift sooner rather than later to avoid confusion and fragmentation. Maintaining configuration in the siloed and unified approaches simultaneously for any longer than necessary is likely to cause problems.

That was my intention. @ryandgoulding @crwr45 Can you guys think of a use case where you want to run different suites against different targets in one run?

jun-rh: By "the rest of the configuration", are you referring to the default labels for auto discovery or the static config schema? My thinking is that the default labels are only useful (persistent) with our test pod. Customer CNF would have to use the custom labels. Similarly, the static config is essentially deprecated and will be used only special cases by dev only. We will make auto discover on by default.

jun-rh: README and config.md updated

jun-rh:
> @crwr45 Thanks, all very good points.

@crwr45 Please take another look.

crwr45 (dismissed):

jun-rh:
Charlie's approval got removed by a new change to fix a typo. Merging it now.

PR #262, opened 2021-06-29
adding logging test : test if PUT emits logs to stdout/stderr

HamadiSE:
Signed-off-by: salaheddine hamadi <shamadi@redhat.com>

### DIFF
--- a/pkg/tnf/handlers/logging/logging.json
+++ b/pkg/tnf/handlers/logging/logging.json
@@ -1,0 +1,30 @@
+{
+  "identifier" : {
+    "url" :  "http://test-network-function.com/tests/logging",
+    "version": "v1.0.0"
+  },
+  "description": "Test if PUT emits output to stdout/stderr",
+  "testResult": 0,
+  "testTimeout": 5000000000,
+  "reelFirstStep": {
+    "execute": "oc logs -n {{.POD_NAMESPACE}} {{.POD_NAME}} {{.CONTAINER_NAME}} --tail 5 | wc -l",
+    "expect":[ "(?m)[1-9]",
+               "(?m)[0]",
+               "(?m)[.*]"],
+    "timeout": 5000000000
+  },
+  "resultContexts":[
+    {
+      "pattern": "(?m)[1-9]",
+      "defaultResult": 1
+    },
+    {
+      "pattern": "(?m)[0]",
+      "defaultResult": 2
+    },
+    {
+      "pattern": "(?m)[.*]",
+      "defaultResult": 0
+    }
+  ]
+}
--- a/pkg/tnf/handlers/logging/logging_test.go
+++ b/pkg/tnf/handlers/logging/logging_test.go
@@ -1,0 +1,146 @@
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package logging_test
+
+import (
+	"fmt"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/generic"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const (
+	testTimeoutDuration = time.Second * 5
+)
+
+var (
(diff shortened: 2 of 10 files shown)


ryandgoulding (commented):
LGTM

ryandgoulding (dismissed):

ryandgoulding:
Should probably update the catalog before merging....

pkg/tnf/handlers/logging/logging_test.go
```diff
@@ -0,0 +1,144 @@
...
+func TestLogging_ReelMatch(t *testing.T) {
+	tester, handlers, jsonParseResult, err := createTest()
+
+	assert.Nil(t, err)
+	assert.True(t, jsonParseResult.Valid())
+	assert.NotNil(t, handlers)
+
+	assert.Equal(t, 1, len(handlers))
+	handler := handlers[0]
+	step := handler.ReelMatch(expectedPassPattern, "", "4")
```
HamadiSE, line 125: Can we unit test the whole matching process ? with ```ReelMatch``` we can only match one pattern at a time.

HamadiSE: Another generic question is :
there's some discrepancy between json schema definition and the code
in tnf/test.go
```
        ERROR = iota
	// SUCCESS represents a successful test.
	SUCCESS
	// FAILURE represents a failed test.
	FAILURE
ERROR = 0; SUCCESS = 1; FAILURE = 2
```
Yet in JSON generic schema
```
0 indicates SUCCESS, 1 indicates FAILURE, 2 indicates ERROR."
```
is this a bug ? or there's a layer that translate the JSON to the right value ?

in all cases, I think we should align the error code between json and the code itself.

pkg/tnf/handlers/logging/logging.json
```diff
@@ -0,0 +1,30 @@
+{
+  "identifier" : {
+    "url" :  "http://test-network-function.com/tests/logging",
+    "version": "v1.0.0"
+  },
+  "description": "Test if PUT emits output to stdout/stderr",
+  "testResult": 0,
+  "testTimeout": 5000000000,
+  "reelFirstStep": {
+    "execute": "oc logs -n {{.POD_NAMESPACE}} {{.POD_NAME}} --tail 5 | wc -l",
```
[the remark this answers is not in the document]
HamadiSE: Right. It should test the container not the pod. In our case, the container is named **test** right ?

jun-rh (approved):

pkg/tnf/handlers/logging/logging_test.go
jun-rh: @ryandgoulding Could you please take a look at Salah's questions?

ryandgoulding: > 0 indicates SUCCESS, 1 indicates FAILURE, 2 indicates ERROR.

The comment is stale.  0 indicates ERROR, 1 indicates SUCCESS, 2 indicates FAILURE.

ryandgoulding: > Can we unit test the whole matching process ? with `ReelMatch` we can only match one pattern at a time.
> For example, in this case, I'm able to test only every pattern alone and can't test the "if else" logic embedded in json definition.

I do not understand what you are asking?  You can test `ReelMatch` from a Unit test context many times, creating a new instance of the Handler/Test each time to reset any localized state.  In each consecutive call, pass the arguments you wish to test.

HamadiSE: my comment is about testing any input and see if it gives the right result at the end. and not testing single pattern against single entry in ReelMatch()

so for example, in my logging.json I have this defined 
```
resultContexts":[
    {
      "pattern": "(?m)[1-9]",
      "defaultResult": 1
    },
    {
      "pattern": "(?m)[0]",
      "defaultResult": 2
    },
    {
      "pattern": "(?m)[.*]",
      "defaultResult": 0
    }
  ]
  ```
  and in the UT I have some tests like -in here, we're testing a single input with a single pattern-
  ```
  handler.ReelMatch(expectedFailPattern, "", "0")
  handler.ReelMatch(expectedErrPattern, "", "a")
  ```
  But I don't have test for the whole logic. By Whole logic, I mean when I send an input, it will return an expected result.
  For example, nothing prevents me from doing something like this :
  ```
resultContexts":[
    {
      "pattern": "(?m)[1-9]",
      "defaultResult": 1
    },
    {
      "pattern": "(?m)[1-9]",
      "defaultResult": 2
    }
  ]
  ```
  if I run both UT, they will pass
   ```
  handler.ReelMatch(expectedPassPattern, "", "0")
  handler.ReelMatch(expectedErrPattern, "", "0")
  ```

what I want, is something like this :

```
input :="a"
assert(my_function("a")==tnf.ERROR)
```
 
not sure if my comment is still unclear. This is my first PR in this project. so there maybe something I'm saying that does not make much sense

Comment (ryandgoulding) [reply pkg/tnf/handlers/logging/logging_test.go:127]:
```
resultContexts":[
  {
    "pattern": "(?m)[1-9]",
    "defaultResult": 1
  },
  {
    "pattern": "(?m)[1-9]",
    "defaultResult": 2
  }
]
```

You should not repeat a pattern regex;  each regex must be distinct to properly capture the output.  Additionally, element order is determined by the ordering in the `*reel.Step.Expect` slice returned  by `ReelFirst()`.  

```
where my_function includes all the logic that parse the input and gives a final result
```

This isn't possible;  `my_function` is bound to the `reel.Handler` at hand (i.e., it is an instance method).  It is not a static function.  You should create a new Handler per test-input.

HamadiSE: Thanks for the explanation.

ryandgoulding:
Requires rebase to resolve merge conflicts.

crwr45 (approved):

ryandgoulding (commented):
Why is `CATALOG.md` deleted here?

PR #269, opened 2021-07-05
Switch to autodiscovery by default

HamadiSE:


jun-rh (commented):
Looks like everything is covered except run-cnf-suite.sh?

HamadiSE:
> Looks like everything is covered except run-cnf-suite.sh?

I did not see anything that needs change in run-cnf-suite.sh. The test does not control the auto-discovery process. Everything is done inside the code using variable environment.

I think we can envisage to add a command line to enable/disable and supply configuration files to the test instead of relying on variable environment. But that's outside the scope of this story.

jun-rh:
> > Looks like everything is covered except run-cnf-suite.sh?
> 
> I did not see anything that needs change in run-cnf-suite.sh. The script does not control the auto-discovery process. Everything is done inside the code using variable environment.
> 
> I think we can envisage to add a command line to enable/disable and supply configuration files to the test instead of relying on variable environment. But that's outside the scope of this story.

Ah, ok, no change is needed because you reversed the logic.

jun-rh (approved):

PR #273, opened 2021-07-16
CNFCERT-56 - Sort and auto update CATALOG.md

edcdavid:


.github/workflows/pre-main.yaml
```diff
@@ -160,3 +160,53 @@ jobs:
...
         if: ${{ github.ref == 'refs/heads/main' && github.repository_owner == 'test-network-function' }}
         run: docker push --all-tags ${REGISTRY}/${IMAGE_NAME}
+ 
+  catalog:
+    name: Update catalog
+    needs: [lint, unit-tests, smoke-tests]
+    runs-on: ubuntu-20.04
+    env:
+      SHELL: /bin/bash
+      KUBECONFIG: '/home/runner/.kube/config'
```
jun-rh, line 170: Not needed?

.github/workflows/pre-main.yaml
```diff
@@ -160,3 +160,53 @@ jobs:
...
+          make build-catalog-md
+      
+      # if there is any changes in the catalog .md file, commit them and schedule a push
+      - name: Commit files
+        id: commit
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "github-actions"
+          git add -f CATALOG.md
+          if [ -z "$(git status --porcelain)" ]; then
```
jun-rh, line 199: Is it better to use a condition based on if identifiers.go is changed? We do that in partner repo, the push to quay.io action is executed only if the docker file is changed.

.github/workflows/pre-main.yaml
edcdavid: Thanks, KUBECONFIG is not needed, will remove.

jun-rh:
I don't see logic to limited this to forked repo. Is the catalog commit gonna be squashed by the merge?

.github/workflows/pre-main.yaml
edcdavid: I would prefer just looking at changes to the Catalog.md file itself because it will catch all scenarios and is very simple. For instance the first time I committed, there was no change in identifers.go but because of a change in sorting, we still needed to commit and push. 
In which file do we check on identifiers.go to make a decision on when to push?

jun-rh: https://github.com/test-network-function/cnf-certification-test-partner/blob/main/.github/workflows/cnf-test-partner-image.yaml

jun-rh: I get your point. We can leave it as is for now and optimize later. The action stuff is free within a limit. We might get a bill if we introduce too much excessive work.

HamadiSE:
looks good to me.

HamadiSE (approved):
looks good to me.

jun-rh:
Having second thought about this whole thing. This step should really happen before the PR is created. Reviewers would prefer to read the formatted md rather than identifiers.go. Feels like we are overthinking. It's not that difficult if we just include it in the makefile as part of default target. If we really **like** github actions, we could do these:

- A status check on the main repo, fail it if identifiers.go is updated and catalog.md is not
- A catalog generate/commit action on forked repo only

edcdavid:
Hi Jun, 
Please see some comments inline:

> I don't see logic to limit this to forked repo only. Is the catalog commit gonna be squashed by the merge?
[David] I did not add any additional triggers, so the pre-main workflow should not be called more than it is now. Pre-main is called before merging to main, on forked branch or upstream, same as before. The only difference is that we rebuild the .md file and push it if needed (any changes). 

> Having second thought about this whole thing. This step should really happen before the PR is created. Reviewers would prefer to read the formatted md rather than identifiers.go.

[David] This is how it works with this PR. The CATALOG.md is updatded when either creating the PR or when pushing to the fork main. In both cases, the CATALOG.md is part of the PR

 Feels like we are overthinking. It's not that difficult if we just include it in the makefile as part of default target. If we really **like** github actions, we could do these:
> 
> * A status check on the main repo, fail it if identifiers.go is updated and catalog.md is not
> * A catalog generate/commit action on forked repo only
> 
[David] Assuming we never check in directly from main repo, the commit will never happen from the main repo. It will always be part of the PR and up for review with just the modified parts.

If we bypass the PR and somehow checkin without a review, the current PR will pickup the changes and update the CATALOG.md

> Making catalog sorted is a really good change. It was impossible to review.

HamadiSE:
> Hi Jun,
> Please see some comments inline:
> 
> > I don't see logic to limit this to forked repo only. Is the catalog commit gonna be squashed by the merge?
> > [David] I did not add any additional triggers, so the pre-main workflow should not be called more than it is now. Pre-main is called before merging to main, on forked branch or upstream, same as before. The only difference is that we rebuild the .md file and push it if needed (any changes).
> 
> > Having second thought about this whole thing. This step should really happen before the PR is created. Reviewers would prefer to read the formatted md rather than identifiers.go.
> 
> [David] This is how it works with this PR. The CATALOG.md is updatded when either creating the PR or when pushing to the fork main. In both cases, the CATALOG.md is part of the PR
> 
> Feels like we are overthinking. It's not that difficult if we just include it in the makefile as part of default target. If we really **like** github actions, we could do these:
> 
> > * A status check on the main repo, fail it if identifiers.go is updated and catalog.md is not
> > * A catalog generate/commit action on forked repo only
> 
> [David] Assuming we never check in directly from main repo, the commit will never happen from the main repo. It will always be part of the PR and up for review with just the modified parts.
> 
> If we bypass the PR and somehow checkin without a review, the current PR will pickup the changes and update the CATALOG.md
> 
> > Making catalog sorted is a really good change. It was impossible to review.

Can we add the action to update the CATALOG.md in the Makefile itself ?
We would guarantee no one misses the generation and we can review the CATALOG and not the identifiers.go 
?

edcdavid:
Hi Salah,
Also talked to Jun with the same comment. I will make the change (add to build-cnf makefile). It will lighten the load in Github and allow developpers to review the .md in case of problems with the auto-generated content

cmd/catalog/cmd/generate.go
```diff
@@ -123,15 +125,28 @@ func emitTextFromFile(filename string) error {
 
 // outputTestCases outputs the Markdown representation for test cases from the catalog to stdout.
 func outputTestCases() {
-	for _, catalogEntry := range identifiers.Catalog {
-		fmt.Fprintf(os.Stdout, "### %s\n", catalogEntry.Identifier.Url)
+
+	//Buildind a separate data structure to store the key order for the map
```
wying3: typo error : Building not Buildind

jun-rh (approved):

PR #274, opened 2021-07-16
Test if containers define the pre-stop hook for lifecycle management

HamadiSE:
Test if containers define the pre-stop hook for lifecycle management
update Dockerfile

Signed-off-by: salah hamadi <shamadi@redhat.com

### DIFF
--- a/pkg/tnf/handlers/shutdown/shutdown.gotemplate
+++ b/pkg/tnf/handlers/shutdown/shutdown.gotemplate
@@ -1,0 +1,3 @@
+{{- range .spec.containers -}}
+  {{ if .lifecycle.preStop }} {{"prestop-defined\n"}} {{- else -}} {{"prestop-not-defined\n"}}{{- end -}}
+{{- end -}}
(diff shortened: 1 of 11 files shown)


pkg/tnf/handlers/shutdown/shutdown.gotemplate
```diff
@@ -0,0 +1,3 @@
+{{- range .spec.containers -}}
```
jun-rh, line 1: The point of this is just to make generic test return something more readable than obscure patterns?

jun-rh: Get it now, it also does a loop. Nice.

jun-rh (approved):

pkg/tnf/handlers/shutdown/shutdown.gotemplate
HamadiSE: Yes exactly. otherwise, it will return the command defined in the output or some "<no value>".

edcdavid: This is nice

edcdavid (approved):
I see the test was failing because the template was not copied in the container

PR #285, opened 2021-07-26
Fixed NON_ZERO regex

greyerof:
The ROOT_CHECK of the access-control TS failed with some valid uid
combinations like:

- 1000640000 --> reported by Ramón Pérez <raperez@redhat.com>
- 1001       --> classic uid used linux/kuberentes examples.
- 001

### DIFF
--- a/pkg/tnf/testcases/base.go
+++ b/pkg/tnf/testcases/base.go
@@ -166,7 +166,7 @@ const (
 	// Zero Allows the result to match 0 number
 	Zero RegExType = "ZERO"
 	// NonZero Allows the result to match non 0 number
-	NonZero RegExType = "NON_ZERO"
+	NonZeroNumber RegExType = "NON_ZERO_NUMBER"
 	// Error Allows the result to match error string
 	Error RegExType = "ERROR"
 	// Digit Allows the result to match any number
@@ -182,7 +182,7 @@ var outRegExp = map[RegExType]string{
 	True:           `^\b(true)\b$`,
 	Null:           `^\b(null)\b$`,
 	Zero:           `0`,
-	NonZero:        `^(0*([1-9]\d*)|null)$`,
+	NonZeroNumber:  `^(0*([1-9]\d*)|null)$`,
 	Error:          "error",
 	Digit:          `\d`,
 }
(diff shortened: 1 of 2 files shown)


ramperher (approved):
Working in local test with pods created with DCI

jun-rh (approved):

pkg/tnf/testcases/base.go
```diff
@@ -182,7 +182,7 @@ var outRegExp = map[RegExType]string{
 	True:           `^\b(true)\b$`,
 	Null:           `^\b(null)\b$`,
 	Zero:           `0`,
-	NonZero:        `^(\d([^0]+)|null)$`,
+	NonZero:        `^(0*([1-9]\d*)|null)$`,
```
HamadiSE, line 185: if your input is ``AA`` this would not match wit ``NonZero``. Is this the intended behavior ? basically, you're always expecting an ``INT`` then testing against it being ``NonZero``

jun-rh: Yes, runasuser is supposed to be a Linux UID

HamadiSE: I'm more concerned with the fact that this ``NonZero`` is defined in ``base.go`` which has a wider scope than the ``runasuser`` test only. 

since you stated this is used in that test only, I'll accept the changes as-is now.

HamadiSE (approved):

pkg/tnf/testcases/base.go
jun-rh: @HamadiSE Ah, yes, good point. This testcases pkg is the other generic test case implementation we talked about during standups. It was developed by Aneesh and used by container/operator tests only. It might be more convenient than the json based one for simple tests. On the other hand, it's probably more limited as well. You can take a closer look if you are interested. I don't see enough reason to keep both approaches. So it's unlikely that we would grow this one.
@greyerof Since we are fixing the pattern anyways, might as well make the name more accurate, maybe NonZeroDigits?

greyerof: @jun-rh Ok, but then I should change the string constant name and value as well. I think they're used in the json template for that TC.

```
	// NonZero Allows the result to match non 0 number
	NonZero RegExType = "NON_ZERO"
```

Btw, why not just "NonZeroNumber" ?

jun-rh: Sure.

PR #288, opened 2021-07-29
Enable lifecycle test suite to run intrusive tests

HamadiSE:


test-network-function/lifecycle/suite.go
```diff
@@ -82,89 +81,163 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
 // All actual test code belongs below here.  Utilities belong above.
 //
 var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
+	configData := common.ConfigurationData{}
+	ginkgo.BeforeSuite(func() {
+		common.Loadconfiguration(&configData)
+		log.Info(configData.ContainersUnderTest)
+	})
+	ginkgo.BeforeEach(func() {
```
HamadiSE, line 89: ```beforeEach``` forces ginkgo to reload configuration before every ``It`` block

test-network-function/common/env.go
```diff
@@ -167,3 +167,52 @@ func NonIntrusive() bool {
...
+func createContainersUnderTest(conf *configsections.TestConfiguration) map[configsections.ContainerIdentifier]*Container {
+	return createContainers(conf.ContainersUnderTest)
+}
+
+// createPartnerContainers sets up the partner containers.
+func createPartnerContainers(conf *configsections.TestConfiguration) map[configsections.ContainerIdentifier]*Container {
+	return createContainers(conf.PartnerContainers)
+}
+
+// Loadconfiguration the configuration into ConfigurationData
```
greyerof, line 190: Loadconfiguration --> LoadConfiguration

test-network-function/lifecycle/suite.go
```diff
@@ -82,89 +81,163 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
...
+	gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
+	gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
+}
+
+func testDeployments(configData *common.ConfigurationData) {
 	var deployments dp.DeploymentMap
 	var notReadyDeployments []string
+	var nodesSorted []node // A slice version of nodes sorted by number of deployments descending
 	ginkgo.When("Testing deployments in namespace", func() {
 		ginkgo.It("Should return list of deployments", func() {
```
jun-rh, line 199: this is no longer accurate

test-network-function/lifecycle/suite.go
```diff
@@ -82,89 +81,163 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
...
+				// We require that all deployments have the desired number of replicas and are all up to date
+				if len(notReadyDeployments) == 0 {
+					ginkgo.Skip("Can not test when deployments are not ready")
+				}
+				gomega.Expect(notReadyDeployments).To(gomega.BeEmpty())
+				ginkgo.By("Should return map of nodes to deployments")
+				nodesSorted = getDeploymentsNodes(namespace)
+				if !common.NonIntrusive() {
+					ginkgo.By("should create new replicas when node is drained")
+					defer results.RecordResult(identifiers.TestPodRecreationIdentifier)
```
jun-rh, line 217: This identifier was removed from the catalog map. Add it back if we are including the test now.

jun-rh:
Question, do we need to add the third node to local test infra for this test to pass?

HamadiSE:
> Question, do we need to add the third node to local test infra for this test to pass?

I'll try it out. I'm using 3 nodes, but I don't expect it to pass.
We have antiAffinity rule, so if we have only two nodes and one node go down, then only one replica would stay up. 
and we will have "Pod number=2"/"Pod Ready = 1"

jun-rh:
> > Question, do we need to add the third node to local test infra for this test to pass?
> 
> I'll try it out. I'm using 3 nodes, but I don't expect it to pass.
> We have antiAffinity rule, so if we have only two nodes and one node go down, then only one replica would stay up.
> and we will have "Pod number=2"/"Pod Ready = 1"

Exactly, so i thought you would add another node first before including it in the test suite.

test-network-function/lifecycle/suite.go
HamadiSE: fixed.

test-network-function/lifecycle/suite.go
HamadiSE: fixed.

HamadiSE:
> > > Question, do we need to add the third node to local test infra for this test to pass?
> > 
> > 
> > I'll try it out. I'm using 3 nodes, but I don't expect it to pass.
> > We have antiAffinity rule, so if we have only two nodes and one node go down, then only one replica would stay up.
> > and we will have "Pod number=2"/"Pod Ready = 1"
> 
> Exactly, so i thought you would add another node first before including it in the test suite. The github workflow tests are passing now because I forgot to update it ;)

New PR is created to have 3 nodes in minikube

test-network-function/lifecycle/suite.go
```diff
@@ -82,90 +81,162 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
...
+		testShutdown(&configData)
 
-		for _, containerUnderTest := range containersUnderTest {
-			testOwner(containerUnderTest.Oc.GetPodNamespace(), containerUnderTest.Oc.GetPodName())
-		}
+		testPodAntiAffinity(&configData)
 
-		for _, containerUnderTest := range containersUnderTest {
-			testPodAntiAffinity(containerUnderTest.Oc.GetPodNamespace())
+		testPodsRecreation(&configData)
```
jun-rh, line 102: Check the non intrusive only flag

HamadiSE: This test can run safely now even in our local machine. That's why I removed the check. I'll add it back

HamadiSE: fixed

test-network-function/lifecycle/suite.go
```diff
@@ -82,90 +81,163 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
...
+	gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
+}
+
+func testPodsRecreation(configData *common.ConfigurationData) {
 	var deployments dp.DeploymentMap
 	var notReadyDeployments []string
-	ginkgo.When("Testing deployments in namespace", func() {
-		ginkgo.It("Should return list of deployments", func() {
+	var nodesSorted []node // A slice version of nodes sorted by number of deployments descending
+	ginkgo.It("Testing node draining effect of deplyment", func() {
```
edcdavid, line 199: deplyment -> deployment

test-network-function/lifecycle/suite.go
```diff
@@ -82,90 +81,163 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
...
+		configData.SetNeedsRefresh()
+		for _, cut := range configData.ContainersUnderTest {
+			namespace := cut.Oc.GetPodNamespace()
+			ginkgo.By(fmt.Sprintf("test deployment in namespace %s", namespace))
 			deployments, notReadyDeployments = getDeployments(namespace)
 			if len(deployments) == 0 {
 				return
 			}
+			// We require that all deployments have the desired number of replicas and are all up to date
+			if len(notReadyDeployments) == 0 {
```
edcdavid, line 209: should't it be !=0 ?

jun-rh (approved):

.github/workflows/pre-main.yaml
```diff
@@ -13,6 +13,7 @@ env:
   IMAGE_TAG: unstable
   TNF_CONTAINER_CLIENT: docker
   TNF_MINIKUBE_ONLY: true
+  TNF_NON_INTRUSIVE_ONLY: true
```
HamadiSE, line 16: updated this line

edcdavid (approved):

edcdavid (approved):

jun-rh (approved):

PR #289, opened 2021-07-29
Use the common.IsMinikube() function inside diagnostic TS.

greyerof:
Local function diagnostic.isMinikube() is a copy of the
common.IsMinikube() one, so it can be deleted.


jun-rh (approved):
Thanks, I missed this suite when promoting the utility function.

HamadiSE (approved):

HamadiSE (approved):

PR #292, opened 2021-08-05
Added scale in/out testing to lifecycle TS

greyerof:
Added new Scaling handler to execute the "oc scale" command. After the
scaling command is issued, the test needs to wait until all deployments
are reported as ready.


pkg/tnf/handlers/scaling/scaling.go
```diff
@@ -0,0 +1,95 @@
...
+func (scaling *Scaling) ReelFirst() *reel.Step {
+	return &reel.Step{
+		Execute: "",
+		Expect:  []string{scaling.regex},
+		Timeout: scaling.timeout,
+	}
+}
+
+// ReelMatch does nothing, just set the test result as success.
+func (scaling *Scaling) ReelMatch(_, _, match string) *reel.Step {
```
HamadiSE, line 83: why it returns success all the time ? so just matching with the "regex" is enough for the test to pass ?

pkg/tnf/handlers/scaling/scaling.go
```diff
@@ -0,0 +1,95 @@
...
+	scaling.result = tnf.SUCCESS
+	return nil
+}
+
+// ReelTimeout does nothing;  no action is necessary upon timeout.
+func (scaling *Scaling) ReelTimeout() *reel.Step {
```
HamadiSE, line 89: do we want here to execute a function to bring back the number of replicas to its original value if things went wrong ?

test-network-function/identifiers/identifiers.go
```diff
@@ -179,6 +179,12 @@ var (
 		Url:     formTestURL(common.LifecycleTestKey, "container-shutdown"),
 		Version: versionOne,
 	}
+
+	// TestScalingIdentifier ensures pre-stop lifecycle is defined
+	TestScalingIdentifier = claim.Identifier{
```
HamadiSE, line 184: update the comment

test-network-function/lifecycle/suite.go
```diff
@@ -104,9 +108,65 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
+
+			if _, alreadyTested := deploymentNames[deploymentName]; alreadyTested {
+				continue
+			}
+
+			// Get deployment's current replicaCount
+			deployments, _ := getDeployments(namespace)
+			replicaCount := deployments[deploymentName].Replicas
+
+			// ScaleIn, removing one pod from the replicaCount
```
HamadiSE, line 146: should you test if `replicas!=1` before scaling down ?
I know that having ``Replicas  == 1`` would fail other tests.. but not sure if we should push replicas count to 0.

test-network-function/lifecycle/suite.go
```diff
@@ -104,9 +108,65 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
+			scaleInHandler := scaling.NewScaling(common.DefaultTimeout, namespace, deploymentName, (replicaCount - 1))
+			testScaleIn, errScaleIn := tnf.NewTest(common.GetContext().GetExpecter(), scaleInHandler, []reel.Handler{scaleInHandler}, common.GetContext().GetErrorChannel())
```
HamadiSE, line 148: ``errScaleIn -> err `` for simplicity, err is used in the next two lines only

pkg/tnf/handlers/scaling/scaling.go
greyerof: Yes. The oc command returns immediately and in case the regex is matched, the command was successful. It only fails in case there's no such deployment name in the cluster.

pkg/tnf/handlers/scaling/scaling.go
greyerof: Good catch.  In my opinion, that restore operation should be done by the testScaling() function, either in case the TC fails by regex or timeout, or because of the notReadyDeployments, done after the scaling, not being zero. What do you think?

test-network-function/lifecycle/suite.go
greyerof: I have tested this manually, and it works. After the "oc scale -n tnf deployment test --replicas=0", you get a 0/0 pods deployment:
```
greyerof@polimnia ~/github/test-network-function (CNFCERT-8)$ oc get deployments -n tnf
NAME      READY   UP-TO-DATE   AVAILABLE   AGE
partner   1/1     1            1           81m
test      0/0     0            0           81m
```

I've launched the scaling TC after a manual scale-in to 1 replicas and the TC works ok, so I don't think that check is necessary here.

test-network-function/lifecycle/suite.go
```diff
@@ -104,9 +108,65 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
 
 		testOwner(&configData)
+
+		testScaling(&configData)
 	}
 })
 
+func waitForAllDeploymentsReady(namespace string, timeout, pollingPeriod time.Duration) {
+	gomega.Eventually(func() []string {
+		_, notReadyDeployments := getDeployments(namespace)
```
jun-rh, line 118: Eventually we should probably replace this with a single deployment query. It might be safe to assume that we own the whole namespace in our lab, but not so much for on-prem test ran by customers.

pkg/tnf/handlers/scaling/scaling.go
HamadiSE: Yes. I'm not really worried about the right place as much as I'm worried about getting things in order if something went off. 
you can even have something like ``defer restoreOriginalCount`` in the ``It`` block
I'm fine with the place you choose.

pkg/tnf/identifier/identifiers.go
```diff
@@ -556,6 +557,18 @@ var Catalog = map[string]TestCatalogEntry{
 			dependencies.OcBinaryName,
 		},
 	},
+	scalingIdentifierURL: {
+		Identifier:  ScalingIdentifier,
+		Description: "A test to check the deployments scale in/out",
```
edcdavid, line 562: Expand the description with a basic explanation of what the test does.

pkg/tnf/handlers/scaling/scaling.go
```diff
@@ -0,0 +1,95 @@
...
+func (scaling *Scaling) Timeout() time.Duration {
+	return scaling.timeout
+}
+
+// Result returns the test result.
+func (scaling *Scaling) Result() int {
+	return scaling.result
+}
+
+// ReelFirst returns a step which expects the ping statistics within the test timeout.
```
edcdavid, line 73: Update the comment

test-network-function/lifecycle/suite.go
```diff
@@ -104,9 +108,65 @@ var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
...
+
+		for _, cut := range configData.ContainersUnderTest {
+			namespace := cut.Oc.GetPodNamespace()
+
+			// Get deployment name and check whether it was already tested.
+			// ToDo: Proper way (helper/handler) to do this.
+			podNameParts := strings.Split(cut.Oc.GetPodName(), "-")
+			deploymentName := podNameParts[0]
+
+			if _, alreadyTested := deploymentNames[deploymentName]; alreadyTested {
```
edcdavid, line 138: where is deploymentNames[] updated with the already tested ?

edcdavid (changes requested):

jun-rh (changes requested):
You updated the test handler identifier only. There is another identifiers.go file under test-network-function directory for test cases.

test-network-function/lifecycle/suite.go
jun-rh: @greyerof Just FYI, we are going to add deploymentUnderTest to the config api soon.

greyerof: @edcdavid It is updated in line 165.
@jun-rh Ok, I'll update it as soon as that new api is available.

test-network-function/identifiers/identifiers.go
greyerof: Done.

test-network-function/lifecycle/suite.go
greyerof: Done.

test-network-function/lifecycle/suite.go
greyerof: Ok.

pkg/tnf/identifier/identifiers.go
greyerof: Done in the new commit.

pkg/tnf/handlers/scaling/scaling.go
greyerof: Done in the new commit.

test-network-function/lifecycle/suite.go
edcdavid: @greyerof ok got it thx

test-network-function/identifiers/identifiers.go
```diff
@@ -461,4 +467,13 @@ the changes for you.`,
...
 			Additionally, ensure that there are available Nodes in the OpenShift cluster that can be utilized in the event that a host Node fails.`,
 	},
+	TestScalingIdentifier: {
+		Identifier: TestScalingIdentifier,
+		Type:       normativeResult,
+		Description: formDescription(TestScalingIdentifier,
+			`tests that CNF deployments support scale in/out operations. 
+			First, The test starts getting the current replicaCount (N) of the deployment/s with the Pod Under Test. Then, it executes the 
+			scale-in oc command for (N-1) replicas. Lastly, it executes the scale-out oc command, restoring the original replicaCount of the deployment/s.`),
+		Remediation: "",
```
jun-rh, line 477: I know there is no real useful remediation we can say. But it's probably not good to leave it empty.

greyerof: Done.

jun-rh (approved):

pkg/tnf/handlers/scaling/scaling.go
greyerof: I've added the deferred restoreDeployments() functions.

HamadiSE (approved):

edcdavid (approved):

PR #293, opened 2021-08-06
Config package refactoring and cleanup

jun-rh:
This is a large PR with real design changes mixed with simple refactoring/renaming changes. Please start reviewing while I work on the README update. Key changes are

- Config schema: sections divided into testTarget and testPartner instead of the old suite based
- Autodiscovery: always on for testPartner section and static config takes precedence
- Autodiscovery: conditional based on the env var for testTarget section. It adds to the static config instead of replacing
- Autodiscovery: custom target pod label support replaces the fixed tnf label based lookup. To keep local test infra working, the default config file has the tnf label in its targetPodLabel list.
- Autodiscovery: "test-network-function.com/container" removed. Pods and containers are discovered by the same set of labels.

Not so important changes are:
- Container -> Pod wherever appropriate
- Cnf -> PodUnderTest
- Eliminated UT packages so that many functions can now be private

Changes pending:
- Readme update: will add to this PR in a subsequent commit
- Updating pod level tests to use the new pod list from the config api (CNFCERT-79)
- Extend discovery process to build a deployment list and update deployment level tests (CNFCERT-80)
- Refactor the test config data structures defined in the test case code base and promote shared ones into pkg/tnf/config (CNFCERT-81)


pkg/config/autodiscover/autodiscover.go
```diff
@@ -67,3 +67,63 @@ func makeGetCommand(resourceType, labelQuery string) *exec.Cmd {
...
+	}
+	for _, c := range containers {
+		containerIDs = append(containerIDs, c.ContainerIdentifier)
+	}
+	return containerIDs, nil
+}
+
+// getContainerByLabel returns exactly one container with the given label. If any other number of containers is found
+// then an error is returned along with an empty `config.Container`.
+func getContainerByLabel(label configsections.Label) (container configsections.Container, err error) {
```
HamadiSE, line 97: question: if we scale a pod to more than one replicas, then the labels will be shared and used by all the replicas.
This function would fail at that point.
in which context we expect to have only one replicas all the time ?

jun-rh: this is used to find the partner containers, i.e. orchestrator and fs-diff, not on targets. This actually could fail with only one pod too if it contains more than 1 containers.

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -0,0 +1,142 @@
...
+package autodiscover
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/test-network-function/test-network-function/pkg/config/configsections"
+	"github.com/test-network-function/test-network-function/pkg/tnf/testcases"
+)
+
+const (
+	configuredTestFile         = "testconfigure.yml"
```
HamadiSE, line 26: testconfigure.yml : does this file need an update ?

jun-rh: No, this file defines the test groups for those combo tests, i.e. host resource test (FKA container) and operator test. I plan to remove the annotation for this in the README first. Eventually we can remove this file too.

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	} else {
+		log.Warnf("an error (%s) occurred when looking for operaters by label", err)
+	}
+
+	return target
+}
+
+// buildPodUnderTest builds a single `configsections.Pod` from a PodResource
+func buildPodUnderTest(pr *PodResource) (cnf configsections.Pod) {
```
HamadiSE, line 73: all those functions with ``build`` and ``create`` names are little bit confusion.
When we should use one or the other ? Also, we're just fetching the data and populating the structure, we're not creating the container/pod itself.
not a comment, but wondering how we should address this

jun-rh: I find them ok, given the context of the autodiscover package. And they are not exported. It would make the names too long if we add some suffix to every type, like podInfo or containerData.

test-network-function/testconfigure.yml
```diff
@@ -16,5 +16,4 @@ operatortest:
   - name: "OPERATOR_STATUS"
     tests:
       - "CSV_INSTALLED"
-      - "SUBSCRIPTION_INSTALLED"
```
edcdavid, line 19: was the subscription_installed test removed on purpose ?

jun-rh: Yes. Close to the end of release 2.0, Ryan created a separate test and removed it from this combo test. The test identifier and this yaml were missed. I updated the identifier in my last PR. Just found out this while going through all of our yml files.

pkg/config/autodiscover/autodiscover_targets.go
jun-rh: Also, I didn't find a function named createXyz in this package. If you are talking about the ones under test-network-function/common, they will be refactored as part of CNFCERT-81. We will make it more consistent then.

HamadiSE: Yes. I was talking in general. and ``buildXyzData`` is more appropriate I guess. ``CreateXyz`` gives the impression that we're creating a ``Pod`` or a ``container``.. but really we're only filling out structures and reading data from the ``deployment``

jun-rh: Sure, those are all in test-network-function/common. Let's align them to a common naming convention once the refactoring is done by CNFCERT-81. I have a feeling that some of the functions may just get combined with others and there won't be as many names to worry about.

README.md
```diff
@@ -14,62 +14,33 @@ certification.  Please see "CNF Developers" below for more information.
...
+- the `CNF under test` is the CNF to be certified. The certification suite identifies the resources (containers/pods/operators etc) belonging to the CNF via labels or static data entries in the config file
 - the `Certification container/exec` is the certification test suite running on the platform or in a container. The executable verifies the CNF under test configuration and its interactions with openshift 
-- the `Partner pod` can be any pod with the required tools (e.g. only ping at the moment) in the same namespace as the `CNF under test`. During connectivity tests, the partner pod will generate pings towards the `CNF under test` to verify connectivity.
-
-These roles are configured in the [Test Configuration](#test-configuration) section below.
+- the `Partner pod` can be any pod with the required tools in the same namespace as the `CNF under test`. For example, during connectivity tests, the partner pod will generate pings towards the `CNF under test` to verify connectivity. The partner pods/containers are auto deployed by the test suite prior a test run and can be auto discovered by the suite without any data entry in the config file.
 
 
 ## Test Configuration
```
edcdavid, line 22: ## Test Configuration

The Test Network Function support auto-configuration using labels and annotations, but also a static configuration using a file. 
The following section describe how to configure the TNF via labels/annotation and the corresponding settings in the config file. A sample config file can be found [here](test-network-function/tnf_config.yml).

It consists of the following sections:
When labelling a pod to be discovered and tested, the discoered pods are **in addition to** the ones
explicitly configured in the testTarget sections of the config file.
### targetPodLabels
The goal of this section is to specify the label to be used to identify the cnf under test pods. So for example, with the default configuration:
```shell script
targetPodLabels:
  - namespace: test-network-function.com
    name: generic
    value: target
```

The corresponding label prefix is: 
```shell script
test-network-function.com/generic: target 
```


### testTarget

README.md
```diff
@@ -14,62 +14,33 @@ certification.  Please see "CNF Developers" below for more information.
...
-```shell
-oc label pod test -n tnf test-network-function.com/generic=target
-oc annotate pod test -n tnf test-network-function.com/skip_connectivity_tests=true
-```
+### targetPodLabels
 
+This section contains a list of labels that will be used to discover the pods and their containers to be tested **in addition to** the ones
+explicitly configured in the testTarget sections.
+### testTarget
+#### podsUnderTest / containersUnderTest
```
edcdavid, line 31: ### podsUnderTest / containersUnderTest

##### Configuration via config file
See below an commented example of adding a single containerand a single pod to test:

```shell script
testTarget:
  containersUnderTest:
    - namespace: tnf # the namespace where the test are run
      podName: test # the name of the pod under test 
      containerName: test the name of the container under test 
      defaultNetworkDevice: eth0 # the name of the default network device (the one assigned by kubernetes core mechanism, not multus)
      multusIpAddresses: # the multus asigned IP
        - 10.217.0.8
  podsUnderTest:
    - name: test # The name of the pod under test
      namespace: tnf # the namespace where the test are run
      tests: # List of tests to run for this pod (List TBD)
        - PRIVILEGED_POD
        - PRIVILEGED_ROLE
```


##### Auto-discovery with labels
Label the pods under test with the target pod label specified in the [targetpodlabels](targetPodLabels) section. 
It's highly recommended that the labels shoud be defined in pod definition rather than added after pod is created, as labels added later on will be lost in case the pod gets rescheduled. In case of pods defined as part a deployment, it's best to use the same label as the one defined in the `spec.selector.matchLabels` section of the deployment yaml.

The autodiscovery mechanism will also attempt to identify the default network device and all the IP addresses of the pods it
needs for network connectivity tests, though that information can be explicitly set using annotations if needed. For Pod IPs:

README.md
```diff
@@ -78,24 +49,15 @@ be seen in [cnf-certification-test-partner](https://github.com/test-network-func
...
-```yaml
-...
-  labels:
-    test-network-function.com/container: target
-  annotations:
-    test-network-function.com/container_tests: "[\"PRIVILEGED_POD\",\"PRIVILEGED_ROLE\"]" # optional
-...
-```
-### Operator Spec
+#### operators
```
edcdavid, line 58: #### operators

##### Configuration via config file
See below an commented example of adding a single operator to test:

```shell script
testTarget:
  operators:
    - name: etcdoperator.v0.9.4 # name of the operator to test
      namespace: default # namespace of the operator to test
      subscriptionName: etcd # Subscription name of the operator
      autogenerate: false # TBD
      tests: # List of tests to run on the operator
        - OPERATOR_STATUS
```
        
##### Auto-discovery with labels

For auto-configuration, do the following:

* CSVs to be tested by the `operator` spec are identified with the `test-network-function-com/operator=target`
label. Any value is permitted but `target` is used here for consistency with the other specs.

README.md
```diff
@@ -104,15 +66,24 @@ annotation. This is equivalent to the `test-network-function.com/container_tests
 * `test-network-function.com/subscription_name` is optional and should contain a JSON-encoded string that's the name of
 the subscription for this CSV. If unset, the CSV name will be used.
 
-### Runtime environement variables to skip tests during development
-#### Turn off openshift required tests when CNF run on kubernetes only environment
-when test on CNFs that run on k8s only environment, execute shell command below before compile tool and run test shell script.
+### testPartner
```
edcdavid, line 69: ### testPartner

This section can also be discovered automatically and should be left commented out unless the parter pods are modified from the original version in [cnf-certification-test-partner](https://github.com/test-network-function/cnf-certification-test-partner/local-test-infra/)

#### partnerContainers
The partner container is used as a partner to initialte connectivity tests
##### Configuration via config file
```shell script
testPartner:
  partnerContainers: # partner containers are listed under this
    - namespace: tnf # the namespace of the partner pod
      podName: partner # the name of the partner pod
      containerName: partner # the container name of the partner pod's container to use
      defaultNetworkDevice: eth0 # the default kubernetes core interface in the pod (Not multus)
      multusIpAddresses: # the multus IP address
        - 10.217.0.29
```

##### Auto-discovery with labels
For auto-configuration, do the add the following label to the pod:
```shell script
test-network-function.com/generic: orchestrator	
```

#### fsDiffMasterContainer
The FSDiffMasterContainer is used as a reference to compare against any modifications in the pod under test during runtime
##### Configuration via config file

```shell script
testPartner:
  fsDiffMasterContainer:
    namespace: tnf # the namespace of the partner pod
    podName: node-master # the name of the pod used to test filesystem
    containerName: master # the name of the container to use in the pod   
```
##### Auto-discovery with labels
For auto-configuration, do the add the following label to the pod:
```shell script
test-network-function.com/generic=fs_diff_master
```



### certifiedcontainerinfo and certifiedoperatorinfo

jun-rh:
A new version of README pushed with some of David's comments.

edcdavid (approved):

HamadiSE (approved):

PR #296, opened 2021-08-09
suites refactoring

HamadiSE:


pkg/tnf/interactive/oc.go
```diff
@@ -40,6 +40,8 @@ type Oc struct {
 	container string
 	// namespace of the pod
 	namespace string
+	// serviceAccountName of the pod
```
jun-rh, line 43: is this the same thing David is adding?

HamadiSE: I did not want to keep passing the serviceAccountName between tests, if you [look here](https://github.com/test-network-function/test-network-function/pull/296/files/58bdf5dc2b0b3c55745d332b8e81b20436cc4eb8#diff-8f2feaab4634f4a165085b1c4b5248d7d792ed5ce4655b5652e354cdf33d3770R183)
serviceAccount is first retrieved, then passed as  an extra param to subsequent tests..
if @edcdavid  is going to clean this up, then I don't mind it being removed

jun-rh: wow, it is indeed super ugly the way it was handled before with the pointer passed into multiple functions. @edcdavid, do you mind replacing this with the one you added when you pick up this change?

test-network-function/certification/suite.go
```diff
@@ -41,43 +41,50 @@ var certAPIClient api.CertAPIClient
...
+		ginkgo.It("get certification status", func() {
+			defer results.RecordResult(identifiers.TestContainerIsCertifiedIdentifier)
 			conf := configpkg.GetConfigInstance()
 			cnfsToQuery := conf.CertifiedContainerInfo
 			if len(cnfsToQuery) > 0 {
 				certAPIClient = api.NewHTTPClient()
-				for _, cnfRequestInfo := range cnfsToQuery {
-					cnf := cnfRequestInfo
+				for _, cnf := range cnfsToQuery {
+					cnf := cnf // pin
```
jun-rh, line 60: Is this change intentional?

test-network-function/lifecycle/suite.go
```diff
@@ -86,10 +86,7 @@ var drainTimeout = time.Duration(drainTimeoutMinutes) * time.Minute
 //
 var _ = ginkgo.Describe(common.LifecycleTestKey, func() {
 	configData := common.ConfigurationData{}
-	ginkgo.BeforeSuite(func() {
-		common.Loadconfiguration(&configData)
-		log.Info(configData.ContainersUnderTest)
-	})
+	configData.SetNeedsRefresh()
```
jun-rh, line 89: Why do we need to force it here?

jun-rh: Oh, I recall the singleton discussion now. We should be able to remove this once the extra layer of needsRefresh is eliminated, right?

test-network-function/platform/suite.go
```diff
@@ -57,68 +57,69 @@ import (
...
-
-		fsDiffTester := cnffsdiff.NewFsDiff(common.DefaultTimeout, containerID)
-		test, err = tnf.NewTest(masterPodOc.GetExpecter(), fsDiffTester, []reel.Handler{fsDiffTester}, masterPodOc.GetErrorChannel())
-		gomega.Expect(err).To(gomega.BeNil())
-		testResult, err = test.Run()
-		gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
-		gomega.Expect(err).To(gomega.BeNil())
+// testPodsFsDiff test that all PUT didn't install new packages are starting
+func testPodsFsDiff(configData *common.ConfigurationData) {
+	ginkgo.It("platform-fsdiff", func() {
```
jun-rh, line 89: You sure we want to remove the existing text? I am ok with adding the prefix.

test-network-function/platform/suite.go
```diff
@@ -57,68 +57,69 @@ import (
...
+				testPodFsDiff(fsDiffContainer.Oc, targetPodOc)
+			}
+		} else {
+			log.Warn("no fs diff container is configured, cannot run fs diff test")
+		}
 	})
 }
 
+// testPodFsDiff  test that the PUT didn't install new packages after starting, and report through Ginkgo.
+func testPodFsDiff(masterPodOc, targetPodOc *interactive.Oc) {
```
jun-rh, line 104: I thought this was a container test. I guess it just test the first one in the pod?

test-network-function/platform/suite.go
HamadiSE: Yes. I can have another story just to go through naming once and for all.. and establish some standard.

test-network-function/certification/suite.go
HamadiSE: Yes. Pinning objects  has something to do with keeping the object in fixed memory address.
My changes keeps the pinning. But , I'm not sure  if it's really necessary.

.github/workflows/pre-main.yaml
```diff
@@ -105,6 +105,7 @@ jobs:
         with:
           repository: test-network-function/cnf-certification-test-partner
           path: cnf-certification-test-partner
+          ref: 350f55ecddbb3b7804b85f444ae023edaf36adff
```
edcdavid, line 108: Should not be needed anymore since  we now check for minikube in the partner repo

test-network-function/generic/suite.go
```diff
@@ -56,50 +55,46 @@ const (
...
-			test, err := tnf.NewTest(oc.GetExpecter(), versionTester, []reel.Handler{versionTester}, oc.GetErrorChannel())
-			gomega.Expect(err).To(gomega.BeNil())
-			testResult, err := test.Run()
-			gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
-			gomega.Expect(err).To(gomega.BeNil())
-		})
+// testIsRedHatRelease fetch the configuration and test containers attached to oc is Red Hat based.
+func testIsRedHatRelease(configData *common.ConfigurationData) {
+	ginkgo.It("Should report a proper Red Hat version", func() {
+		for _, cut := range configData.ContainersUnderTest {
```
edcdavid, line 77: just now realized what "cut" means :-)  => container under test

edcdavid (approved):

test-network-function/lifecycle/suite.go
HamadiSE: Yes. Once all the config is collapsed, we'll work with singleton and expose a single API to update everything.

edcdavid (approved):

jun-rh (changes requested):
testFsDiff is a container level test.

jun-rh (approved):

edcdavid (approved):

PR #297, opened 2021-08-09
CNFCERT-73

isaacdorfman:
Added sysctlAllConfigsArgs identifier.


pkg/tnf/dependencies/binaries.go
```diff
@@ -50,6 +50,9 @@ const (
 	// OcBinaryName is the name of the OpenShift CLI client command.
 	OcBinaryName = "oc"
 
+	// SysctlBinaryName is the name of the Sysctl command.
+	SysctlBinaryName = "oc"
```
jun-rh, line 54: Is this intentional or it should be a different command name?

jun-rh (changes requested):
You updated the handler catalog, not the test case catalog. See the other identifiers.go under test-netowrk-function.

pkliczewski:
@isaacdorfman please rebase

jun-rh (approved):

HamadiSE (approved):

PR #304, opened 2021-08-12
CNFCERT-30 changes

isaacdorfman:


run-cnf-suites.sh
```diff
@@ -66,5 +66,12 @@ if [[ ! -f "/proc/1/cgroup" ]] || grep -q init\.scope /proc/1/cgroup; then
 	cd ..
 fi
 
+if [[ -z "${TNF_PARTNER_SRC_DIR}" ]]; then
```
jun-rh, line 69: This breaks github workflow. Maybe the github workflow should align with how developers typically do this? i.e. checkout partner repo and run this script with the env var set, instead of invoking the github action from the other repo directly?

run-cnf-suites.sh
```diff
@@ -66,5 +66,12 @@ if [[ ! -f "/proc/1/cgroup" ]] || grep -q init\.scope /proc/1/cgroup; then
...
+	echo "env var \"TNF_PARTNER_SRC_DIR\" not set, please set it before running this command!"
+	exit 1
+fi
+
+make -C $TNF_PARTNER_SRC_DIR install-partner-pods
```
edcdavid, line 74: This is breaking the container TNF image. Make and other tools are removed before the image is finished. Need to test and update testing using container image

edcdavid (changes requested):

run-cnf-suites.sh
isaacdorfman: @jun-rh @edcdavid I fixed this problem, I added this env var to pre-main.yaml.

jun-rh:
Do you mind updating line 48 in pre-main.yaml? We missed it when shuffling the tests.

run-cnf-suites.sh
isaacdorfman: @edcdavid @pkliczewski 
"Need to test and update testing using container image"
You mean adding a test to the CI that uses the the image built in the previous tests.

isaacdorfman:
@jun-rh @edcdavid "make" is no longer removed in from the image.

run-cnf-suites.sh
jun-rh: @test-network-function/cnfcert Team, please comment if you have better suggestion than this mandatory env var.

jun-rh: For now, I am ok with making this (partner pods installation) run in the container case only, so that the customer case is still covered and no extra step introduced for developers.

jun-rh (approved):

HamadiSE (approved):

edcdavid (approved):

PR #305, opened 2021-08-13
Singleton test env/config structure

jun-rh:


pkg/config/config.go
```diff
@@ -20,25 +20,28 @@ import (
...
 
 var (
-	// configInstance is the singleton instance of loaded config, accessed through GetConfigInstance
-	configInstance configsections.TestConfiguration
-	// loaded tracks if the config has been loaded to prevent it being reloaded.
-	loaded = false
-	// set when an intrusive test has done something that would cause Pod/Container to be recreated
-	needsRefresh = false
+	// testEnvironment is the singleton instance of `TestEnvironment`, accessed through `GetTestEnvironment`
+	testEnvironment TestEnvironment
```
HamadiSE, line 44: In here, by using ``TestEnvironment`` instead of ``testEnvironment``, users can actually instantiate new instances of this. 
by not doing it, we can't have function that takes the singleton as parameter, so each function need to actually call ``config.GetInstance()`` which is okay solution, but requires more work :) 
WDYT ?

pkg/config/config.go
```diff
@@ -62,41 +138,77 @@ func loadConfigFromFile(filePath string) error {
...
 	if autodiscover.PerformAutoDiscovery() {
-		configInstance.TestTarget = autodiscover.FindTestTarget(configInstance.TargetPodLabels)
-		autodiscover.FillTestPartner(&configInstance.TestPartner)
+		autodiscover.FindTestTarget(env.Config.TargetPodLabels, &env.Config.TestTarget)
+	}
+	autodiscover.FindTestPartner(&env.Config.Partner)
+
+	log.Infof("Test Configuration: %+v", *env)
+
+	for _, cid := range env.Config.ExcludeContainersFromConnectivityTests {
```
HamadiSE, line 174: what does cid stands for here ?

jun-rh: i think container id

HamadiSE (approved):

edcdavid (approved):

PR #309, opened 2021-08-17
List all installed CSI drivers info from cluster CNFCERT24-item 1

wying3:
add listing 3rd party CSI drivers info to diagnostic test

### DIFF
--- a/test-network-function/diagnostic/suite.go
+++ b/test-network-function/diagnostic/suite.go
@@ -35,8 +35,17 @@ var (
 
 	nodesHwInfo = NodesHwInfo{}
 
+	// csiDriver stores the csi driver JSON output of `oc get csidriver -o json`
+	csiDriver = CsiDriver{}
+
 	// nodesTestPath is the file location of the nodes.json test case relative to the project root.
 	nodesTestPath = path.Join("pkg", "tnf", "handlers", "node", "nodes.json")
+
+	// csiDriverTestPath is the file location of the csidriver.json test case relative to the project root.
+	csiDriverTestPath = path.Join("pkg", "tnf", "handlers", "csidriver", "csidriver.json")
+
+	// relativeCsiDriverTestPath is the relative path to the csidriver.json test case.
+	relativeCsiDriverTestPath = path.Join(pathRelativeToRoot, csiDriverTestPath)
 
 	// pathRelativeToRoot is used to calculate relative filepaths for the `test-network-function` executable entrypoint.
 	pathRelativeToRoot = path.Join("..")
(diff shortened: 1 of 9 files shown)


jun-rh:
Can you put a real example of the claim file section of this in a comment?

wying3:
> Can you put a real example of the claim file section of this in a comment?

kind: CSIDriver
metadata:
  labels:
    app: controller.csi.trident.netapp.io
name: csi.trident.netapp.io
selfLink: /apis/storage.k8s.io/v1/csidrivers/csi.trident.netapp.io

HamadiSE (dismissed):

test-network-function/diagnostic/suite.go
```diff
@@ -112,6 +125,18 @@ type NodesHwInfo struct {
 	Worker NodeHwInfo
 }
 
+// CsiDriver stores csi driver JSON output from  `oc get csidirver -o json` command
+type CsiDriver struct {
+	Kind     string `json:"kind"`
```
jun-rh, line 130: what is the point of this field if it's always the same value?

test-network-function/diagnostic/suite.go
```diff
@@ -112,6 +125,18 @@ type NodesHwInfo struct {
...
+	Metadata struct {
+		Labels struct {
```
jun-rh, line 132: Should this be an array or list?

test-network-function/diagnostic/suite.go
wying3: just want to specifiy in the claim file that this part is CSI Driver info

jun-rh:
Took a look at the output Varun provided. There seems to be a lot more info that could be potentially useful, like version, group police etc. Why don't we just capture everything as {}interface type, similar to the node summary, instead of debating what fields to pick and creating structures for them?

test-network-function/diagnostic/suite.go
wying3: changed to map list

test-network-function/diagnostic/suite.go
```diff
@@ -112,6 +125,18 @@ type NodesHwInfo struct {
...
+			App string `json:"app"`
```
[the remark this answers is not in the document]
wying3: changed to new data  structure to include all labels

wying3:
> Took a look at the output Varun provided. There seems to be a lot more info that could be potentially useful, like version, group police etc. Why don't we just capture everything as {}interface type, similar to the node summary, instead of debating what fields to pick and creating structures for them?

changed made with new commit

jun-rh (approved):

HamadiSE (approved):

PR #312, opened 2021-08-18
update tests naming

HamadiSE:
enable focus and skip specific tests


test-network-function/accesscontrol/suite.go
```diff
@@ -73,7 +73,8 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 
 //nolint:gocritic,funlen // ignore hugeParam error. Pointers to loop iterator vars are bad and `testCmd` is likely to be such.
 func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase, testType string) {
-	ginkgo.It(fmt.Sprintf("Running Pod test : %s", testCmd.Name), func() {
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestHostResourceIdentifier)
```
edcdavid, line 76: testCmd.Name is required here for the It name. All access-control sub tests have the same name otherwise

test-network-function/operator/suite.go
```diff
@@ -189,18 +191,17 @@ func agrName(operatorName, subName, testName string) string {
...
-			opInTest := operator.NewOperator(cmdArgs, name, namespace, testCmd.ExpectedStatus, testCmd.ResultType, testCmd.Action, defaultTimeout)
-			gomega.Expect(opInTest).ToNot(gomega.BeNil())
-			test, err := tnf.NewTest(context.GetExpecter(), opInTest, []reel.Handler{opInTest}, context.GetErrorChannel())
-			gomega.Expect(err).To(gomega.BeNil())
-			gomega.Expect(test).ToNot(gomega.BeNil())
-			testResult, err := test.Run()
-			gomega.Expect(err).To(gomega.BeNil())
-			gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
-		})
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestOperatorInstallStatusIdentifier)
```
edcdavid, line 194: Same here, I believe we need to use testCmd.Name instead of identifiers.TestOperatorInstallStatusIdentifier

edcdavid (changes requested):
see comments inside

test-network-function/accesscontrol/suite.go
HamadiSE: fixed

test-network-function/operator/suite.go
HamadiSE: fixed

run-tnf-container.sh
```diff
@@ -54,13 +56,13 @@ usage() {
 	     (currently: $HOME/.kube/config).
 
 	Examples
-	  $0 -t ~/tnf/config -o ~/tnf/output diagnostic generic
+	  $0 -t ~/tnf/config -o ~/tnf/output -f diagnostic generic -s diagnostic-nodes-hw-info
```
jun-rh, line 59: replace generic

run-tnf-container.sh
```diff
@@ -54,13 +56,13 @@ usage() {
...
 
 	  Because -k is omitted, $(basename $0) will first try to autodiscover local kubeconfig files.
 	  If it succeeds, the diagnostic and generic tests will be run using the autodiscovered configuration.
 	  The test results will be saved to the '~/tnf/output' directory on the host.
 
-	  $0 -k ~/.kube/ABC:~/.kube/DEF -t ~/tnf/config -o ~/tnf/output diagnostic generic
+	  $0 -k ~/.kube/ABC:~/.kube/DEF -t ~/tnf/config -o ~/tnf/output -f diagnostic generic
```
jun-rh, line 65: same as above

jun-rh:
Do you have a junit.xml generated from this? Easier to review the real output.

run-tnf-container.sh
HamadiSE: fixed.

run-tnf-container.sh
HamadiSE: fixed

HamadiSE:
> Do you have a junit.xml generated from this? Easier to review the real output.

I pushed the xml files so you can take a look, I'll remove them later
The used command is 
```
reset; ./run-cnf-suites.sh -f networking lifecycle -s lifecyle-container-shutdown
```

test-network-function/lifecycle/suite.go
```diff
@@ -425,7 +430,9 @@ func uncordonNode(node string) {
 func testPodAntiAffinity(env *config.TestEnvironment) {
 	var deployments dp.DeploymentMap
 	ginkgo.When("CNF is designed in high availability mode ", func() {
-		ginkgo.It("Should set pod replica number greater than 1 and corresponding pod anti-affinity rules in deployment", func() {
+		testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestPodHighAvailabilityBestPractices)
+		ginkgo.It(testID, func() {
+			ginkgo.By("Should set pod replica number greater than 1 and corresponding pod anti-affinity rules in deployment")
```
jun-rh, line 435: "By" doesn't really show. Same as my comment on the other PR, I don't see why the original text has to be removed. If the new addition has to be at the end, maybe add a prefix so it's sort of separated. How about:

ginkgo.It(fmt.Sprintf("Should set pod replica number greater than 1 and corresponding pod anti-affinity rules in deployment. Test Id: %s", identifiers.XformToGinkgoItIdentifier(identifiers.TestPodHighAvailabilityBestPractices)), func(){

jun-rh: Or maybe put it into when, but at the front? That way we don't need to repeat suite name.

HamadiSE: this is used in xml file only. is there really a need to write this very long description in the test?
we have that in identifiers. By  will only show up if you have verbose.
in the script run-cnf-test; we need to pass -gingko.v

edcdavid (approved):

test-network-function/lifecycle/suite.go
greyerof: Seems like the By text will only appear if the test fails. As per ginkgo [doc](https://onsi.github.io/ginkgo/#documenting-complex-its-by):
`The string passed to By is emitted via the GinkgoWriter. If a test succeeds you won’t see any output beyond Ginkgo’s green dot. If a test fails, however, you will see each step printed out up to the step immediately preceding the failure. Running with ginkgo -v always emits all steps.`

HamadiSE: @jun-rh  is on PTO.
@edcdavid @greyerof do you think I can leave it as-is and pass the -ginko.v all the times ?
or I should update the test identifier to include the description as well ?

greyerof: I'd do both things:
1. Use jun's It text version (first jun's comment). I like ginkgo's/bdd descriptive text for the It blocks.
2. pass always the -ginkgo.v.  This could create large logs, but can be disabled at any moment.

HamadiSE: one of the main objective for this change is to make it easier to identify tests and allow users to skip tests.
here's a log when running the program
```
./run-cnf-suites.sh -f lifecycle
...
...
lifecycle 
  lifecycle-scaling
  /Users/shamadi/redhat/go/test-network-function/test-network-function/lifecycle/suite.go:169
INFO[0005] Testing deployment=test, namespace=tnf pod name=test-697ff58f87-hqxdm 
2021/08/20 14:23:50 Sent: "oc -n tnf get deployments -o custom-columns=NAME:.metadata.name,REPLICAS:.spec.replicas,READY:.status.readyReplicas,UPDATED:.status.updatedReplicas,AVAILABLE:.status.availableReplicas,UNAVAILABLE:.status.unavailableReplicas && echo END_OF_TEST_SENTINEL\n"
2021/08/20 14:23:50 Match for RE: "(?s).+((.|\n)*END_OF_TEST_SENTINEL\n)" found: ["NAME      REPLICAS   READY   UPDATED   AVAILABLE   UNAVAILABLE\npartner   1          1       1         1           <none>\ntest      2          2       2         2           <none>\nEND_OF_TEST_SENTINEL\n" "END_OF_TEST_SENTINEL\n" ""] Buffer: "NAME      REPLICAS   READY   UPDATED   AVAILABLE   UNAVAILABLE\npartner   1          1       1         1           <none>\ntest      2          2       2         2           <none>\nEND_OF
```
now, the user can easily detect the suite and the test name. 
then he can use the command 
```
./run-cnf-suites.sh -f lifecycle -s lifecycle-scaling
``` 
the test description is not unique and can mess up the regular expressions.
If we want to add it to When; that should be fine, the only drawback is that it will appear on JUNIT
```
      </testcase>
      <testcase name="lifecycle lifecycle-pod-scheduling" classname="CNF Certification Test Suite" time="2.901942578"></testcase>
      <testcase name="lifecycle when Test terminationGracePeriod  lifecycle-pod-termination-grace-period" classname="CNF Certification Test Suite" time="0.202937519"></testcase>
      <testcase name="lifecycle when Testing PUTs are configured with pre-stop lifecycle lifecycle-container-shutdown" classname="CNF Certification Test Suite" time="0.220083457"></testcase>
      <testcase name="lifecycle when CNF is designed in high availability mode  lifecycle-pod-high-availability" classname="CNF Certification Test Suite" time="0.421853337"></testcase>
      <testcase name="lifecycle lifecycle-pod-recreation" classname="CNF Certification Test Suite" time="0">
          <skipped></skipped>
      </testcase>
```
see the difference between lifecycle-pod-scheduling and the others...

if something went bad, then we can rely on the logs to give us more data

wying3: test name list  .e.g.  lifecycle-scaling etc,  should be added into README.md, so user can pick name from list to -s to cancel test

test-network-function/accesscontrol/suite.go
```diff
@@ -73,7 +73,8 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 
 //nolint:gocritic,funlen // ignore hugeParam error. Pointers to loop iterator vars are bad and `testCmd` is likely to be such.
 func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase, testType string) {
-	ginkgo.It(fmt.Sprintf("Running Pod test : %s", testCmd.Name), func() {
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestHostResourceIdentifier) + "-" + testType
```
edcdavid, line 76: Did we loose  changes here? I previously commented on this, we should use testCmd.Name instead of testType

HamadiSE: not sure what happened. I fixed it.
The reason I want to also delay that ID fix is because I'm not sure how to handle this tests here.

HamadiSE: do we have those tests in the catalog.md btw ?

edcdavid (approved):
looks good to me

test-network-function/accesscontrol/suite.go
edcdavid: yes,  but they are all clumped under the host-resource identifier.

run-tnf-container.sh
```diff
@@ -21,12 +21,14 @@ export REQUIRED_VARS_ERROR_MESSAGES=(
 	'OUTPUT_LOC is required. Use the -o option to specify the output location for the test results.'
 )
 
-export TNF_IMAGE_NAME=test-network-function
-export TNF_IMAGE_TAG=latest
+export TNF_IMAGE_NAME=salah #test-network-function
```
wying3, line 24: why TNF_IMAGE_NAME change to your name and TAG is not latest

edcdavid (approved):

wying3 (dismissed):

edcdavid (approved):

wying3 (approved):

edcdavid (approved):

PR #314, opened 2021-08-19
Shamadi node draning fix

HamadiSE:


wying3 (approved):
tested in both clusters, e.g., cnfdt12 and cnfcert 3nodes cluster , but pass the test.

test-network-function/lifecycle/suite.go
```diff
@@ -314,36 +314,36 @@ func testPodsRecreation(env *config.TestEnvironment) {
 			if len(deployments) == 0 {
 				return
 			}
+			if _, exists := namespaces[podNamespace]; !exists {
+				namespaces[podNamespace] = true
+			}
```
greyerof, line 319: You'll repeat the getDeploymentsNode(podNamespace) on line 326 for each pod in the same namespace. Is it necessary? I suggest adding here this:
```
else {
    continue
}
```

HamadiSE: right. will fix it

wying3 (approved):
reapprove

greyerof (approved):

PR #317, opened 2021-08-23
Fix for HostPath test 

edcdavid:
Using Go template to to look for hostpath volumes. If <no value> is reported one or more times, the test passes. Otherwise the test fails.

### DIFF
--- a/pkg/tnf/testcases/data/cnf/privilegedpod.go
+++ b/pkg/tnf/testcases/data/cnf/privilegedpod.go
@@ -43,10 +43,10 @@ var PrivilegedPodJSON = string(`{
       "name": "HOST_PATH_CHECK",
       "skiptest": true,
        "loop": 0,
-      "command": "oc get pod  %s  -n %s -o json  | jq -r '.spec.hostpath.path'",
+      "command": "oc get pods %s -n %s -o go-template='{{ range .spec.volumes}}{{.hostPath.path}}{{end}}'",
       "action": "allow",
       "expectedstatus": [
-        "NULL_FALSE"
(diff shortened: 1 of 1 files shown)


pkg/tnf/testcases/data/cnf/privilegedpod.go
```diff
@@ -43,10 +43,10 @@ var PrivilegedPodJSON = string(`{
       "name": "HOST_PATH_CHECK",
       "skiptest": true,
        "loop": 0,
-      "command": "oc get pod  %s  -n %s -o json  | jq -r '.spec.hostpath.path'",
+      "command": "oc get pods %s -n %s -o go-template='{{ range .spec.volumes}}{{.hostPath.path}}{{end}}'",
       "action": "allow",
       "expectedstatus": [
-        "NULL_FALSE"
+        "^(<no value>)*$"
       ]
```
HamadiSE, line 50: you're expecting nothing in here for this test to pass ?
that seems a weird test. I though you need to find volumes mounted over a network

edcdavid: After looking at the initial test again, this was really the intent, the implementation was just flawed. The function I tested last week was also looping through the PVC and PVs, but is beyond
 what we intend to test before and for this we will need a template file.

wying3: same question, why not find host-path mounted

HamadiSE (approved):

wying3 (approved):
verified in a pod with host-path defined. fix works.

PR #328, opened 2021-08-29
update_operator

shimritproj:


jun-rh (dismissed):

test-network-function/identifiers/identifiers.go
```diff
@@ -323,7 +323,9 @@ the same hacks.'`),
 		Description: formDescription(TestOperatorInstallStatusIdentifier,
 			`Ensures that CNF Operators abide by best practices.  The following is tested:
 1. The Operator CSV reports "Installed" status.
-2. TODO: Describe operator scc check.`),
+2. Control the actions that a pod can perform and what it has the ability to access,
```
wying3: maybe change from  what it has the ability to access to capabilities it has to access

shimritproj: maybe I can to adding : In this test we check whether clusterPermissions are set for specific
ClusterServiceVersion is installed. We test operator which is not
using the permissions but this tests enable us to check whether operator
in test is using privileged and in this case we should fail and
analyze why such permission is needed.

edcdavid: Ying and I met about this, we propose the following:
Check that the operator is not configured with privileged rights. If clusterPermissions exists in the CSV, and the clusterPermissions[].rules[].resourceNames is not empty then we fail the test. Otherwise, the test passes. In case the test fails, we need to analyze the need for the privilege mode.

CATALOG.md
```diff
@@ -179,7 +179,7 @@ Suggested Remediation|Ensure that your Operator is installed via OLM.
 Property|Description
 ---|---
 Version|v1.0.0
-Description|http://test-network-function.com/testcases/operator/install-status Ensures that CNF Operators abide by best practices.  The following is tested: 1. The Operator CSV reports "Installed" status. 2. TODO: Describe operator scc check.
+Description|http://test-network-function.com/testcases/operator/install-status Ensures that CNF Operators abide by best practices.  The following is tested: 1. The Operator CSV reports "Installed" status. 2. Controls the actions that pods created by the operator can perform (e.g. checks that the operator is not configured with privileged rights). If clusterPermissions exists in the CSV, and can perform the clusterPermissions[]. rules[].resourceNames is not empty then we fail the test. Otherwise, the test passes. In case the test fails, we need to analyze the need for the privilege mode.
```
wying3, line 182: the word "can perform" in front of clusterPermissions[] can be removed

edcdavid (approved):

CATALOG.md
jun-rh: Not sure if describing the exact logic in English is a good idea. Seems pretty hard to read to me.

jun-rh: I see the recommendation from David and Ying. It reads better than the current version which is a strange mix of Shimrit's original with David and Ying's text.

wying3 (approved):

test-network-function/identifiers/identifiers.go
```diff
@@ -323,7 +323,10 @@ the same hacks.'`),
 		Description: formDescription(TestOperatorInstallStatusIdentifier,
 			`Ensures that CNF Operators abide by best practices.  The following is tested:
 1. The Operator CSV reports "Installed" status.
-2. TODO: Describe operator scc check.`),
+2. Check that the operator is not configured with privileged rights.
+If clusterPermissions exists in the CSV,
+and the clusterPermissions[].rules[].resourceNames is not empty then we fail the test.
+Otherwise, the test passes. In case the test fails, we need to analyze the need for the privilege mode.`),
```
HamadiSE, line 329: In case the test fails, we need to analyze the need for the privilege mode.
if the test fail and you have recommendation, then it should go in remediation part.
I did not like writing exactly what the test is doing using If else, this should be a general description of the test and what it aims to check, not how it does the check exactly.

jun-rh: I made the same point in the other comment. Also, avoid using "we" as it feels informal.

jun-rh: I made the same point in other comment. Also, avoid using "we". It feels informal. Describe it in an objective way if we need to, e.g.: Test passes if [clusterPermissions is not present in the CSV manifest or is present with no resourceNames under its rules].
Also agree on Salah's comment on the remediation part. In this case though, it's fairly obvious what should be checked. I am actually ok with the existing generic remediation text. Just remove the last sentence.

shimritproj: Ok, So I will change the description -  

Check that the operator is not configured with privileged rights.
If clusterPermissions exists in the CSV,
and the clusterPermissions[].rules[].resourceNames is not empty then fail the test.
Otherwise, the test passes.

it's OK?

jun-rh: As discussed in the team channel, I am updating the text so this can be part of 3.1.

jun-rh (approved):

edcdavid (approved):

PR #334, opened 2021-09-01
Update folders to check in fsdiff

HamadiSE:


jun-rh (approved):

pkg/tnf/handlers/cnffsdiff/cnffsdiff.go
```diff
@@ -73,14 +78,22 @@ func (p *CnfFsDiff) ReelFirst() *reel.Step {
 func (p *CnfFsDiff) ReelMatch(pattern, before, match string) *reel.Step {
 	p.result = tnf.SUCCESS
 	switch pattern {
-	case lib:
```
edcdavid, line 76: we do not have a case for "lib"? I see the regex  defined above

edcdavid (approved):

pkg/tnf/handlers/cnffsdiff/cnffsdiff.go
HamadiSE: fixed

jun-rh (approved):

PR #345, opened 2021-09-07
Expose cnf feature deploy test versoin and skip arg through env vars

jun-rh:


script/run-cfd-container.sh
```diff
@@ -1,11 +1,12 @@
...
 
-if [ "$VERIFY_CNF_FEATURES" == "true" ] && [ "$TNF_MINIKUBE_ONLY" != "true" ]; then
+if [ "$TNF_RUN_CFD_TEST" == "true" ] && [ "$TNF_MINIKUBE_ONLY" != "true" ]; then
 
 	export TNF_IMAGE_NAME=cnf-tests
-	export TNF_IMAGE_TAG=4.6
+	export TNF_IMAGE_TAG="${TNF_CFD_IMAGE_TAG:-4.6}"
 	export TNF_OFFICIAL_ORG=quay.io/openshift-kni/
 	export TNF_OFFICIAL_IMAGE="${TNF_OFFICIAL_ORG}${TNF_IMAGE_NAME}:${TNF_IMAGE_TAG}"
+	export TNF_CFD_SKIP="${TNF_CFD_SKIP:-performance|sriov|ptp|sctp|xt_u32|dpdk|ovn}"
```
HamadiSE, line 9: do we want to skip those tests all the time ?
also, by default, if you check this line : 
https://github.com/test-network-function/test-network-function/blob/main/script/run-container.sh#L115
you only need to define TNF_SKIP_SUITES

jun-rh: > do we want to skip those tests all the time ?
> also, by default, if you check this line :
> https://github.com/test-network-function/test-network-function/blob/main/script/run-container.sh#L115
> you only need to define TNF_SKIP_SUITES

cfd image takes raw ginkgo skip arg, not the one we use with our run-cnf-suites.sh.

HamadiSE: we're still using ./run-container at the end. it's not like we're running the binary from this script.

I'll pull your PR and give it a try

HamadiSE: I did run the test. I did not even know we're calling other repositories to test something. 
I was able to run the tests properly from my machine.
right now, it's running 2 out of 17 specs. and the way the script is built will prevent us from enabling the sriov tests for example.

``` TNF_CFD_SKIP="${TNF_CFD_SKIP:-performance|sriov|ptp|sctp|xt_u32|dpdk|ovn}" ```
means the sriov.. is always skipped. I think we should have another FLAG that indicates which tests should pass and we filter out the remaining

Comment (jun-rh) [reply script/run-cfd-container.sh:9]: > I did run the test. I did not even know we're calling other repositories to test something.
> I was able to run the tests properly from my machine.
> right now, it's running 2 out of 17 specs. and the way the script is built will prevent us from enabling the sriov tests for example.
> 
> `TNF_CFD_SKIP="${TNF_CFD_SKIP:-performance|sriov|ptp|sctp|xt_u32|dpdk|ovn}"`
> means the sriov.. is always skipped. I think we should have another FLAG that indicates which tests should pass and we filter out the remaining

No it doesn't do that:

```
$ echo $X

$ export X=${X:-x}
$ echo $X
x
$ export X=${X:-y}
$ echo $X
x
```

HamadiSE (approved):

wying3 (approved):

PR #349, opened 2021-09-08
Fix for operator tests passing with no operator

edcdavid:


test-network-function/operator/suite.go
```diff
@@ -70,6 +69,7 @@ var _ = ginkgo.Describe(testSpecName, func() {
 		env := config.GetTestEnvironment()
 		ginkgo.BeforeEach(func() {
 			env.LoadAndRefresh()
+			gomega.Expect(len(env.OperatorsUnderTest)).ToNot(gomega.Equal(0))
```
jun-rh, line 72: The intent is to fail the test if there is no operator found?

edcdavid: yes

wying3 (approved):

test-network-function/operator/suite.go
HamadiSE: should we not just skip it ? Is operator deployment mandatory ?

jun-rh: exactly, similar to what we did in the scaling and node drain test, skip when no deployments are found.

edcdavid: done

HamadiSE (approved):

PR #353, opened 2021-09-13
Replace exec.Command with Expecter

shimritproj:


pkg/tnf/handlers/command/command.go
```diff
@@ -43,7 +43,7 @@ const (
 func NewCommand(timeout time.Duration, resourceType, labelQuery, namespace string) *Command {
 	command := fmt.Sprintf(ocCommand, resourceType, namespace, labelQuery)
 	return &Command{
-		timeout: timeout,
+		timeout: 2 * timeout,
```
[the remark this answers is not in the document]
shimritproj: I was comeback the variable timeout what has before. this working when I running the command "./run-cnf-suites.sh -f lifecycle" and when I running this command " " make build-cnf-tests && TNF_MINIKUBE_ONLY=true TNF_NON_INTRUSIVE_ONLY=true ./run-cnf-suites.sh -f lifecycle".

what do you think?

pkg/tnf/handlers/command/command.go
```diff
@@ -0,0 +1,98 @@
...
+type Command struct {
+	result  int
+	timeout time.Duration
+	args    []string
+	Output  string
+	// adding special parameters
+}
+
+const (
+	ocCommand = "oc get %s -n %s -o json -l %s"
```
jun-rh, line 39: Shouldn't the handler named OcGetCommand if "oc get" is hardcoded? If it meant to be generic, the command part should become an arg as well.

pkg/tnf/dependencies/binaries.go
```diff
@@ -29,6 +29,9 @@ const (
 	// CutBinaryName is the name of the Unix `cut` command.
 	CutBinaryName = "cut"
 
+	// CommandBinaryName is the name of the Unix `command` command.
+	CommandBinaryName = "command"
```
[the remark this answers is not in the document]
shimritproj: Yes, all commands others did it like this.
for example: 	
// HostnameBinaryName is the name of the Unix `hostname` command.
	HostnameBinaryName = "hostname"

	// IPBinaryName is the name of the Unix `ip` command.
	IPBinaryName = "ip"

	// LsBinaryName is the name of the Unix `ls` command.
	LsBinaryName = "ls"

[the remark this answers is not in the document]
greyerof: But those other handlers actually need those existing unix/linux commands to work. Your handler relied only in "oc", which was already in that list.

pkg/tnf/handlers/command/command.go
```diff
@@ -0,0 +1,89 @@
...
+type Command struct {
+	result  int
+	timeout time.Duration
+	args    []string
+	Output  string
+	// adding special parameters
+}
+
+// NewGetCommand creates a new command handler.
+func NewGetCommand(timeout time.Duration, command string) *Command {
```
greyerof, line 38: Change this function's name to NewCommand().

pkg/tnf/handlers/command/command.go
```diff
@@ -0,0 +1,89 @@
...
+func (h *Command) Timeout() time.Duration {
+	return h.timeout
+}
+
+// Result returns the test result.
+func (h *Command) Result() int {
+	return h.result
+}
+
+// ReelFirst returns a step which expects an Command summary for the given device.
```
greyerof, line 66: Please update comment. The part about the "summary for the given device" does not apply.

pkg/tnf/identifier/identifiers.go
```diff
@@ -611,6 +612,12 @@ var Catalog = map[string]TestCatalogEntry{
 	},
 }
 
+// CommandIdentifier is  the Identifier used to represent the generic command test case.
+var CommandIdentifier = Identifier{
```
jun-rh, line 616: This should be added into the catalog map. Give it a good description that explains what use case this handler is for, i.e. can be used with any command and would match any output. The caller is responsible for interpreting the output and extracting data from it.

PR #356, opened 2021-09-16
add default timeout to send command in spawned shell

hamadise:
This aims to add timeout to "Send" function in goexpect library, by default, the library does not use any timeout. 
By looking in the code of the goexpect,  it's printing the log after "sending" the command, so if there's any issue in the "send" we will not see the log. This seems like the issue we were seeing this morning where the process hangs.

### DIFF
--- a/cmd/generic/cmd/jsontest.go
+++ b/cmd/generic/cmd/jsontest.go
@@ -251,7 +251,7 @@ func runOcCmd(_ *cobra.Command, args []string) {
 	// oc shell creation.
 	goExpectSpawner := interactive.NewGoExpectSpawner()
 	var spawnContext interactive.Spawner = goExpectSpawner
-	oc, ch, err := interactive.SpawnOc(&spawnContext, pod, container, namespace, (*tester).Timeout(), interactive.Verbose(true))
+	oc, ch, err := interactive.SpawnOc(&spawnContext, pod, container, namespace, (*tester).Timeout(), interactive.Verbose(true), interactive.SendTimeout((*tester).Timeout()))
 	if err != nil {
 		fatalError("could not create the oc expecter", err, testExpecterCreationFailedExitCode)
 	}
--- a/cmd/generic/cmd/jsontest.go
+++ b/cmd/generic/cmd/jsontest.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"time"
 
 	expect "github.com/google/goexpect"
 	"github.com/google/goterm/term"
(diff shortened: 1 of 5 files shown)


test-network-function/common/env.go
```diff
@@ -44,7 +44,7 @@ var DefaultTimeout = time.Duration(defaultTimeoutSeconds) * time.Second
 
 // GetContext spawns a new shell session and returns its context
 func GetContext() *interactive.Context {
-	context, err := interactive.SpawnShell(interactive.CreateGoExpectSpawner(), DefaultTimeout, interactive.Verbose(true))
+	context, err := interactive.SpawnShell(interactive.CreateGoExpectSpawner(), DefaultTimeout, interactive.Verbose(true), interactive.SendTimeout(DefaultTimeout*time.Second))
```
greyerof, line 47: What about the other two contexts (SpawnOc/SpawnSSH)?

greyerof: I think the "*time.Second" is not needed here, since it was already done to the DefaultTimeout var.

pkg/tnf/interactive/spawner.go
```diff
@@ -146,6 +146,11 @@ type GoExpectSpawner struct {
 	verboseWriterIsSet bool
 	// verboseWriter is an alternate destination for verbose logs.
 	verboseWriter io.Writer
+
+	// sendTimeoutIsSet tracks whether the Send command timeout is set.
+	sendTimeoutIsSet bool
+	// sendTimeout is the timeout of send command
+	sendTimeout time.Duration
```
greyerof, line 153: I think the GetGoExpectOptions() function needs to be updated to include this option in the return array.

test-network-function/common/env.go
hamadise: right. I'll fix it now

pkg/tnf/interactive/spawner.go
hamadise: Thanks for catching it.

greyerof (approved):

cmd/generic/cmd/jsontest.go
```diff
@@ -224,7 +224,7 @@ func runSSHCmd(_ *cobra.Command, args []string) {
 	// SSH shell creation.
 	goExpectSpawner := interactive.NewGoExpectSpawner()
 	var spawnContext interactive.Spawner = goExpectSpawner
-	context, err := interactive.SpawnSSH(&spawnContext, user, host, (*tester).Timeout(), interactive.Verbose(true))
+	context, err := interactive.SpawnSSH(&spawnContext, user, host, (*tester).Timeout(), interactive.Verbose(true), interactive.SendTimeout((*tester).Timeout()))
 	if err != nil {
 		fatalError("could not create the ssh expecter", err, testExpecterCreationFailedExitCode)
```
edcdavid, line 229: Line 254: should runOcCmd use the extra sendtimeout option? 
oc, ch, err := interactive.SpawnOc(&spawnContext, pod, container, namespace, (*tester).Timeout(), interactive.Verbose(true))

hamadise: fixed.

edcdavid (approved):

PR #357, opened 2021-09-16
Spawn session error handling

jun-rh:


crwr45 (approved):
Good catch, this looks OK now and should work as intended.

Reading back over the code I think the previous error was a typo (`outCh` on line 83 should have been `chOut`) that was not caught by linting due to the `:=`. That would have meant the scope of `chOut` was such the previous code would have worked.

crwr45 (approved):

greyerof (approved):

hamadise (approved):

PR #362, opened 2021-09-21
Adding stderr to logger

edcdavid:
reading the stderr pipe of spawned commands and displaying as a warning in logger

### DIFF
--- a/pkg/tnf/interactive/spawner.go
+++ b/pkg/tnf/interactive/spawner.go
@@ -18,6 +18,7 @@ package interactive
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
@@ -325,7 +326,7 @@ func (g *GoExpectSpawner) Spawn(command string, args []string, timeout time.Dura
 		return nil, err
 	}
 
-	cmdLine := strings.Join(args, " ")
+	cmdLine := fmt.Sprintf("%s %s", command, strings.Join(args, " "))
 	logCmdMirrorPipe(cmdLine, stderrPipe, "STDERR", false)
 	stdoutPipe = logCmdMirrorPipe(cmdLine, stdoutPipe, "STDOUT", true)
 
--- a/pkg/tnf/interactive/spawner.go
+++ b/pkg/tnf/interactive/spawner.go
@@ -294,7 +294,6 @@ func LogStdErr(cmdLine string, stderrpipe io.Reader) {
 			log.Warn(err)
 			return
 		}
-		time.Sleep(100 * time.Millisecond) //nolint:gomnd //100 ms variable used only here
 	}
 }
 
--- a/pkg/tnf/interactive/spawner.go
+++ b/pkg/tnf/interactive/spawner.go
@@ -282,7 +282,7 @@ func NewGoExpectSpawner() *GoExpectSpawner {
 }
 
 // LogStdErr logs stderr output to logger (warning)
-func LogStdErr(cmdLine string, stderrpipe io.Reader) {
+func logStdErr(cmdLine string, stderrpipe io.Reader) {
 	buf := bufio.NewReader(stderrpipe)
 
 	for {
@@ -317,7 +317,7 @@ func (g *GoExpectSpawner) Spawn(command string, args []string, timeout time.Dura
 	}
 
 	cmdLine := strings.Join(args, " ")
-	go LogStdErr(cmdLine, stderrPipe)
+	go logStdErr(cmdLine, stderrPipe)
 
 	err = g.startCommand(spawnFunc, command, args)
 	if err != nil {
--- a/pkg/tnf/interactive/spawner.go
+++ b/pkg/tnf/interactive/spawner.go
@@ -282,7 +282,7 @@ func NewGoExpectSpawner() *GoExpectSpawner {
 	return &GoExpectSpawner{}
 }
 
-// LogStdErr logs stderr output to logger (warning)
+// logCmdMirrorPipe logs specified pipe output to logger.
 func logCmdMirrorPipe(cmdLine string, pipeToMirror io.Reader, name string, trace bool) io.Reader {
 	originalPipe := pipeToMirror
 	r, w, _ := os.Pipe()
--- a/pkg/tnf/interactive/spawner.go
+++ b/pkg/tnf/interactive/spawner.go
@@ -17,10 +17,12 @@
 package interactive
 
 import (
+	"bufio"
 	"io"
 	"os"
 	"os/exec"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/tnf/interactive/spawner.go
```diff
@@ -272,6 +281,23 @@ func NewGoExpectSpawner() *GoExpectSpawner {
 	return &GoExpectSpawner{}
 }
 
+// LogStdErr logs stderr output to logger (warning)
+func LogStdErr(cmdLine string, stderrpipe io.Reader) {
+	buf := bufio.NewReader(stderrpipe)
+
+	for {
+		line, _, err := buf.ReadLine()
+		log.Warn("STDERR for " + cmdLine + " : " + string(line))
```
hamadise, line 290: why not using logrus.Error ?

pkg/tnf/interactive/spawner.go
```diff
@@ -272,6 +281,23 @@ func NewGoExpectSpawner() *GoExpectSpawner {
...
+
+		if err != nil {
+			// Some Error has happened, goroutine about to exit
+			log.Warn(err)
```
hamadise, line 294: ditto

pkg/tnf/interactive/spawner.go
edcdavid: the output of stderr might or might not be an error for our main program. With warning I was trying to say that it could be a error

edcdavid: for instance:
WARN[0072] STDERR for exec -n tnf -it test-697ff58f87-c8vcv -c test -- sh : Unable to use a TTY - input is not a terminal or the right kind of file

greyerof: I also prefer Warning here.

pkg/tnf/interactive/spawner.go
greyerof: I think this is actually an error so I would go with log.Error here.

pkg/tnf/interactive/spawner.go
```diff
@@ -272,6 +281,23 @@ func NewGoExpectSpawner() *GoExpectSpawner {
...
+			return
+		}
+		time.Sleep(100 * time.Millisecond) //nolint:gomnd //100 ms variable used only here
```
greyerof, line 297: Is this sleep really necessary? The buf.ReadLine() seems to be a blocking call.

pkg/tnf/interactive/spawner.go
edcdavid: This is how we exit this go routine for instance when the discovery recompute the container under test. There is an error reading the pipe. It is not really an error execpt maybe in certain cases

greyerof: You're right.

pkg/tnf/interactive/spawner.go
edcdavid: fixed

pkg/tnf/interactive/spawner.go
```diff
@@ -272,6 +281,22 @@ func NewGoExpectSpawner() *GoExpectSpawner {
 	return &GoExpectSpawner{}
 }
 
+// LogStdErr logs stderr output to logger (warning)
+func LogStdErr(cmdLine string, stderrpipe io.Reader) {
```
hamadise, line 285: do we need to export this function outside of spawner ? I think using "logStdErr" instead of "LogStdErr" is preferred unless we want to use it somewhere else ?

edcdavid: fixed

pkg/tnf/interactive/spawner.go
```diff
@@ -286,10 +311,14 @@ func (g *GoExpectSpawner) Spawn(command string, args []string, timeout time.Dura
 	}
 
 	spawnFunc = (*spawnFunc).Command(command, args...)
-	stdinPipe, stdoutPipe, err := g.unpackPipes(spawnFunc)
+	stdinPipe, stdoutPipe, stderrPipe, err := g.unpackPipes(spawnFunc)
 	if err != nil {
 		return nil, err
 	}
+
+	cmdLine := strings.Join(args, " ")
```
greyerof, line 319: You're not using the Spawn()'s "command" argument, which is usually (but not always) "oc". I suggest changing this for something like:
cmdLine := fmt.Sprintf("%s %s", command, strings.Join(args, " "))

edcdavid: done

jun-rh: I thought args already included the command?

	// Args holds command line arguments, including the command as Args[0].
	// If the Args field is empty or nil, Run uses {Path}.
	//
	// In typical use, both Path and Args are set by calling Command.
	Args []string

pkg/tnf/interactive/spawner.go
```diff
@@ -272,6 +282,31 @@ func NewGoExpectSpawner() *GoExpectSpawner {
 	return &GoExpectSpawner{}
 }
 
+// LogStdErr logs stderr output to logger (warning)
+func logCmdMirrorPipe(cmdLine string, pipeToMirror io.Reader, name string, trace bool) io.Reader {
```
hamadise, line 286: comment does not match function name

pkg/tnf/interactive/spawner.go
```diff
@@ -272,6 +282,31 @@ func NewGoExpectSpawner() *GoExpectSpawner {
...
+	go func() {
+		for {
+			buf := bufio.NewReader(tr)
+			line, _, err := buf.ReadLine()
+			if trace {
+				log.Trace(name + " for " + cmdLine + " : " + string(line))
+			} else {
+				log.Warn(name + " for " + cmdLine + " : " + string(line))
+			}
+			if err != nil {
```
hamadise, line 300: what's the rational behind logging either via Trace or Warn ?
I think logging everytime in trace level is good.  Also, you're always passing "true" and there's not a way to change that via command line, so it's really trace everytime.

edcdavid: This was to address your comment about not having the stdout at warn level. leave stderr at warn and set stdout at trace

pkg/tnf/interactive/spawner.go
jun-rh: Ah, your change is in the setup phase where command and args are still separate as function args. I thought you were getting it from the SpawnFunc interface which wraps os.Cmd.Args().

jun-rh (approved):

pkg/tnf/interactive/spawner.go
edcdavid: fixed

hamadise (approved):

PR #363, opened 2021-09-22
Oc session error channel monitoring enhancement for intrusive test

jun-rh:


pkg/config/config.go
```diff
@@ -84,8 +84,15 @@ func getOcSession(pod, container, namespace string, timeout time.Duration, optio
...
 		go func() {
-			err := <-outCh
-			gomega.Expect(err).To(gomega.BeNil())
+			for {
+				select {
+				case err := <-outCh:
+					log.Errorf("SpawnOc error: %v from %s %s", err, oc.GetPodName(), oc.GetPodContainerName())
+					gomega.Expect(err).To(gomega.BeNil())
+				case <-oc.GetDoneChannel():
+					break
```
edcdavid, line 93: Shouldn't we kill the cmd process here as well ?

edcdavid (approved):

pkg/config/config.go
jun-rh: We could if we really want to release resources a little bit sooner. Probably not for now.

pkg/config/config.go
```diff
@@ -84,8 +84,15 @@ func getOcSession(pod, container, namespace string, timeout time.Duration, optio
 		gomega.Expect(err).To(gomega.BeNil())
 		// Set up a go routine which reads from the error channel
 		go func() {
-			err := <-outCh
-			gomega.Expect(err).To(gomega.BeNil())
+			for {
+				select {
+				case err := <-outCh:
+					log.Errorf("SpawnOc error: %v from %s %s", err, oc.GetPodName(), oc.GetPodContainerName())
+					gomega.Expect(err).To(gomega.BeNil())
```
greyerof, line 91: Needs a previous defer ginkgo.GinkgoRecover() to avoid panic'ing in case err is not nil.

jun-rh: Hmm, I was wondering whether the original assertion is really the appropriate action here. Recover would let the test run continue, right? I don't think that's right either. Is there an alternative to panic if we just want to stop the test run?

hamadise: gingko.Fail() will only fail the test and continue, unless you have fast fail, in which case, the test would stop on the first failure.

test-network-function/lifecycle/suite.go
```diff
@@ -275,54 +275,54 @@ func shutdownTest(podNamespace, podName string) {
...
+		ginkgo.By(fmt.Sprintf("test deployment in namespace %s", env.NameSpaceUnderTest))
+		deployments, notReadyDeployments = getDeployments(env.NameSpaceUnderTest)
+		if len(deployments) == 0 {
+			return
 		}
+		// We require that all deployments have the desired number of replicas and are all up to date
+		if len(notReadyDeployments) != 0 {
+			ginkgo.Skip("Can not test when deployments are not ready")
+		}
+		gomega.Expect(notReadyDeployments).To(gomega.BeEmpty())
```
hamadise, line 293: line 293 code is not useful.. based on line 291, this test will always pass.

hamadise: it can be removed safely

pkg/tnf/interactive/oc.go
```diff
@@ -109,3 +111,13 @@ func (o *Oc) GetOptions() []Option {
 func (o *Oc) GetErrorChannel() <-chan error {
 	return o.errorChannel
 }
+
+// GetDoneChannel returns the receive only done channel
+func (o *Oc) GetDoneChannel() <-chan bool {
```
hamadise, line 116: This function is not needed.. 
The moment you receive something in the channel, you can return in getOcSession function (line 92)
```
case <- oc :
break
```

jun-rh: huh? oc is not a channel.

pkg/config/config.go
jun-rh: Panic is the only way then? Personally I don't mind. It just looks a bit scary to people who don't know what it means. I could improve the log text a bit so they understand that the panic is intentional?

pkg/tnf/interactive/oc.go
hamadise: my bad. I thought you're getting the status of the channel and not the channel itself. And I was suggesting to get test directly against the channel. anyway, ignore this comment.

pkg/config/config.go
hamadise: I'm not sure what are you trying to achieve here, but here's what to expect :
 * a failure in gomega or ginkgo.Fail would result in current test (ginkgo.It ) be stopped and the code will move to next test.
 * panic will stop everything.

jun-rh: I think we should stop the test altogether, as the test result wouldn't be useful if we continue running with broken sessions.

greyerof: Problem is this code is not part of any test: it's run inside the autodiscovery mechanism in the BeforeEach section. So, if the oc shell cannot be created, even using the defer ginkgo.GinkgoRecover() will make the first TC in the test suite and those after any intrusive TC to fail.
I'd go with removing the gomega assertion and use a logrus.Warn/Error message. These ocSessions are only used for the connectivity TCs, right?

jun-rh: No, they are used by other tests too.

hamadise (approved):

greyerof (approved):

PR #365, opened 2021-09-24
Get OCP version and save it in claim file

edcdavid:
I will remove the temporary library call to my repo once the schema repo is checked in


pkg/tnf/handlers/versionocp/version.go
```diff
@@ -0,0 +1,115 @@
+// Copyright (C) 2021 Red Hat, Inc.
```
hamadise, line 1: I would prefer clusterversion.go as a name of the file and package as version is very generic term.

test-network-function/suite_test.go
```diff
@@ -164,9 +164,13 @@ func getTNFVersion() *version.Version {
 }
 
 // incorporateTNFVersion adds the TNF version to the claim.
-func incorporateTNFVersion(claimData *claim.Claim) {
+func incorporateVersions(claimData *claim.Claim) {
 	claimData.Versions = &claim.Versions{
-		Tnf: getTNFVersion().Tag,
+		Tnf:      getTNFVersion().Tag,
+		TnfGitCommit: GitCommit,
+		OcClient: diagnostic.GetVersionsOcp()[0],
```
hamadise, line 171: It would be better if you have :
```
Ocp:  diagnostic.GetClusterVersion().Ocp
```
with cluster version struct has the following fields 
```
ocp string
oc string
k8s string
```
what do you think ?

test-network-function/diagnostic/suite.go
```diff
@@ -199,6 +215,18 @@ func listNodeCniPlugins(nodeName string) []CniPlugin {
...
 
+func testOcpVersion() {
+	context := common.GetContext()
+	tester := versionocp.NewVersionOCP(defaultTestTimeout)
+	test, err := tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
+	gomega.Expect(err).To(gomega.BeNil())
+	testResult, err := test.Run()
+	gomega.Expect(testResult).To(gomega.Equal(tnf.SUCCESS))
+	gomega.Expect(err).To(gomega.BeNil())
+	versionsOcp = tester.GetVersions()
```
hamadise, line 226: I think it's better if we have more explicit GetVersions return instead of []strings

pkg/tnf/handlers/versionocp/version.go
edcdavid: done

test-network-function/suite_test.go
edcdavid: done

test-network-function/diagnostic/suite.go
edcdavid: done

hamadise (approved):

test-network-function/identifiers/identifiers.go
```diff
@@ -200,6 +200,11 @@ var (
 		Url:     formTestURL(common.DiagnosticTestKey, "cluster-csi-info"),
 		Version: versionOne,
 	}
+	// TestclusterVersionIdentifier list Cluster CSIdriver Identifier retrieves Third Party CSI driver info.
+	TestclusterVersionIdentifier = claim.Identifier{
```
jun-rh, line 204: We don't add diagnostic test cases to catalog?

edcdavid: yes, we should, the catalog was updated

jun-rh: only for the handler, not for the test cases. need to add this into the map.

edcdavid: fixed

pkg/tnf/identifier/identifiers.go
```diff
@@ -596,6 +597,18 @@ var Catalog = map[string]TestCatalogEntry{
 			dependencies.OcBinaryName,
 		},
 	},
+	clusterVersionIdentifierURL: {
+		Identifier:  ClusterVersionIdentifier,
+		Description: "extracts OCP versions from the cluster",
```
greyerof, line 602: Extracts

edcdavid: fixed

PR #372, opened 2021-09-29
New TC: CRDs status subresource availability

greyerof:
Iterates through target CRDs to and checks whether they have a valid
status subresource definition.

### DIFF
--- a/pkg/tnf/handlers/crdstatus/crdstatus.go
+++ b/pkg/tnf/handlers/crdstatus/crdstatus.go
@@ -1,0 +1,129 @@
+// Copyright (C) 2020-2021 Red Hat, Inc.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package crdstatus
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/dependencies"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	ocLabelQueryFlag = " -l "
+	ocCommand        = "%s get crds -o json"
+	jqCommand        = "'[ .items[] | {\"name\": .metadata.name, \"kind\" : .spec.names.kind, \"status\" : .spec.versions[].schema.openAPIV3Schema.properties.status}]'"
+)
+
+// Crd maps to a custom json object to get the Status config of a CRD.
+type Crd struct {
+	Name   string                 `json:"name"`
+	Kind   string                 `json:"kind"`
+	Status map[string]interface{} `json:"status"`
+}
+
+// CrdStatus provides a CrdStatus test implemented using command line tool CrdStatus.
+type CrdStatus struct {
+	result  int
+	timeout time.Duration
+	args    []string
+
+	CrdItems []Crd
+	// adding special parameters
+}
+
+func getCommand(labels []string) string {
+	command := fmt.Sprintf(ocCommand, dependencies.OcBinaryName)
+
+	if len(labels) > 0 {
+		command += ocLabelQueryFlag
+		for _, label := range labels {
+			command += label + ","
+		}
+		// Remove last ","
+		command = command[:len(command)-1]
+	}
+
+	// Add jq filter.
+	command += " | " + dependencies.JqBinaryName + " " + jqCommand
+	return command
+}
+
+// NewCrdStatus returns a new CrdStatus test handler.
+func NewCrdStatus(timeout time.Duration, labels []string) *CrdStatus {
+	return &CrdStatus{
+		result:   tnf.ERROR,
+		timeout:  timeout,
+		args:     strings.Split(getCommand(labels), " "),
+		CrdItems: []Crd{},
+	}
+}
+
+// Args returns the handler arguments
+func (h *CrdStatus) Args() []string {
+	return h.args
+}
+
+// GetIdentifier returns the tnf.Test specific identifier.
+func (h *CrdStatus) GetIdentifier() identifier.Identifier {
+	return identifier.DeploymentsIdentifier
+}
+
+// Timeout return the timeout for the test.
+func (h *CrdStatus) Timeout() time.Duration {
+	return h.timeout
+}
+
+// Result returns the test result.
+func (h *CrdStatus) Result() int {
+	return h.result
+}
+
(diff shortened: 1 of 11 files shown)


pkg/tnf/handlers/crdstatus/crdstatus.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/sirupsen/logrus"
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/dependencies"
+	"github.com/test-network-function/test-network-function/pkg/tnf/identifier"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+)
+
+const (
+	ocLabelQueryFlag = " -l "
```
jun-rh, line 33: I don't think CRDs would typical share the same labels as the target pods. For now, I am ok with a namespace-wide query without labels. If we really want to allow them to target specific CRDs, it should be a separate config, not in the targetPodLabel list.

greyerof: With "namespace-wide" query you mean adding the "-n <namespace>" flag to the "oc get crds" command? That wouldn't work, because crds seem not to be namespaced. Their spec says whether the its CR objects can be namespaced or cluster-wide, but the CRD itself is always cluster wide. So I can see two options:
a) Use the "-l" to filter CRDs with custom labels for CRDs (different from the pod ones that I'm using in my PR) as you've suggested.
b) 1. Get a list of all CRDs of the cluster. 2. Check if there are CR objects of that CRD in the CNF target namespace. If so, fail if the that specific CRD has no status spec.

pkg/tnf/handlers/crdstatus/crdstatus.go
```diff
@@ -0,0 +1,129 @@
...
+	return command
+}
+
+// NewCrdStatus returns a new CrdStatus test handler.
+func NewCrdStatus(timeout time.Duration, labels []string) *CrdStatus {
+	return &CrdStatus{
+		result:   tnf.ERROR,
+		timeout:  timeout,
+		args:     strings.Split(getCommand(labels), " "),
+		CrdItems: []Crd{},
```
hamadise, line 78: ``` args : string[]{getCommand()} ```
you don't need to build the command, then split it, then pass it to Reel to be joined again
you can just pass the full command as one thing.

pkg/tnf/handlers/crdstatus/crdstatus.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Args returns the handler arguments
+func (h *CrdStatus) Args() []string {
+	return h.args
+}
+
+// GetIdentifier returns the tnf.Test specific identifier.
+func (h *CrdStatus) GetIdentifier() identifier.Identifier {
+	return identifier.DeploymentsIdentifier
```
hamadise, line 89: is this intentional ? I think we should have separate identifier for this.

pkg/tnf/handlers/crdstatus/crdstatus.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Result returns the test result.
+func (h *CrdStatus) Result() int {
+	return h.result
+}
+
+// ReelFirst returns a step which expects an CrdStatus summary for the given device.
+func (h *CrdStatus) ReelFirst() *reel.Step {
+	return &reel.Step{
+		Expect:  []string{".+"}, // TODO : pass the list of possible regex in here
```
hamadise, line 105: do we want this ```(?ms).+``` single line matching ? The current mode you're using will match only the first line normally!

pkg/tnf/handlers/crdstatus/crdstatus.go
```diff
@@ -0,0 +1,129 @@
...
+		Timeout: h.timeout,
+	}
+}
+
+// ReelMatch parses the CrdStatus output and set the test result on match.
+func (h *CrdStatus) ReelMatch(_, _, match string) *reel.Step {
```
hamadise, line 111: as general, anytime we don't need complex logic in the ReelMatch, we should use .json
This case is good candidate for Json.
But the work is already done. I just want to point this out.

pkg/tnf/handlers/crdstatus/crdstatus_test.go
```diff
@@ -0,0 +1,86 @@
...
+	assert.NotNil(t, handler)
+	handler.ReelEOF()
+}
+
+func Test_ReelTimeout(t *testing.T) {
+	handler := NewCrdStatus(testTimeoutDuration, testCrdTargetLabels)
+	assert.NotNil(t, handler)
+	assert.Nil(t, handler.ReelTimeout())
+}
+
```
hamadise, line 77: can you move the const and var section up ?
it's better to see variables declared before being used. I kept scrolling to find them :)

pkg/tnf/identifier/identifiers.go
```diff
@@ -59,6 +59,7 @@ const (
 	scalingIdentifierURL                  = "http://test-network-function.com/tests/scaling"
 	csiDriverIdentifierURL                = "http://test-network-function.com/tests/csiDriver"
 	clusterVersionIdentifierURL           = "http://test-network-function.com/tests/clusterVersion"
+	crdStatusIdentifierURL                = "http://test-network-function.com/tests/crdStatus"
```
hamadise, line 62: I see now. This is the identifier that should be returned.

test-network-function/diagnostic/suite.go
```diff
@@ -67,6 +70,9 @@ var (
 
 var _ = ginkgo.Describe(common.DiagnosticTestKey, func() {
 	if testcases.IsInFocus(ginkgoconfig.GinkgoConfig.FocusStrings, common.DiagnosticTestKey) {
+		env := config.GetTestEnvironment()
+		env.LoadAndRefresh()
+
```
hamadise, line 75:
```
env := config.GetTestEnvironment()
		ginkgo.BeforeEach(func() {
			env.LoadAndRefresh()
			gomega.Expect(len(env.PodsUnderTest)).ToNot(gomega.Equal(0))
			gomega.Expect(len(env.ContainersUnderTest)).ToNot(gomega.Equal(0))
		})
```
we mad add some intrusive test in the future to this section (highly unlikely, but for the sake of completeness)

test-network-function/diagnostic/suite.go
```diff
@@ -355,3 +367,29 @@ func listClusterCSIInfo() {
...
+		ginkgo.Skip("No CRDs found using target labels.")
+	}
+
+	for _, crd := range handler.CrdItems {
+		ginkgo.By("Checking CRD: " + crd.Name)
+		if crd.Status == nil {
+			ginkgo.Fail("CRD " + crd.Name + " does not have status spec.")
+		}
+	}
+}
```
hamadise, line 395: few questions here :
do we want this to be in the diagnostic suite ? if yes, do we want to add any of this information to the claim file ?

pkg/tnf/handlers/crdstatus/crdstatus.go
greyerof: I know that, but I just (blindly) followed the Args() interface function, which expects an array of strings as the returned value.

pkg/tnf/handlers/crdstatus/crdstatus.go
greyerof: Yet somehow it works. The whole multi-line json object string is set as "match" parameter in the ReelMatch function and it's correctly unmarshalled.

pkg/tnf/handlers/crdstatus/crdstatus.go
jun-rh: ah, i see. maybe use spec.group instead of forcing them to label CRDs one by one. For option b, i worry if the CRs don't always exist.

test-network-function/diagnostic/suite.go
jun-rh: Salah is right. This shouldn't be in diagnostic. It was the "dumping all CRD" alternative that should be part of diagostic. I guess this one could fall into "obsevability"?

PR #373, opened 2021-09-29
CNFCERT-92:Extend auto discovery process to build a node list and upd…

aabughosh:
…ate node level tests


pkg/config/configsections/Nodes.go
```diff
@@ -0,0 +1,23 @@
...
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package configsections
+
+// Nodes defines in the cluster. with name of the node and the type of this node master/worker,,,,,
+type Nodes struct {
+	Name string
+	Type []string
```
hamadise, line 22: I think we have only two types, master and worker
you can use this 
```
type Node struct {
	// Name is the name of node
	Name string `yaml:"name" json:"name"`
	// Worker indicates if the node is defined as worker in OCP cluster
	Worker bool `yaml:"worker" json:"worker"`
	// Master indicates if the node is defined as master in OCP cluster
	Master bool `yaml:"master" json:"master"`
}
```
Also, this Node not Nodes.

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,55 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+				Type: append(types, "worker"),
+			}
+			for j := range nodes {
+				if nodeNames[i] == nodes[j].Name {
+					nodes[j].Type = append(nodes[j].Type, "worker")
+					Existinarray = 1
+				}
+			}
+			if Existinarray == 0 {
+				nodes = append(nodes, node)
```
hamadise, line 117: ```Existinarray``` should be set back to 0 
otherwise, you'll be always adding nodes to nodes

Also, I believe it's better to have a map

pkg/config/config.go
```diff
@@ -199,6 +201,7 @@ func (env *TestEnvironment) doAutodiscover() {
 	env.TestOrchestrator = env.PartnerContainers[env.Config.Partner.TestOrchestratorID]
 	env.DeploymentsUnderTest = env.Config.DeploymentsUnderTest
 	env.OperatorsUnderTest = env.Config.Operators
+	env.Nodes = env.Config.Nodes
```
hamadise, line 204: it's better to have env.Nodes as a map[string]*Node

pkg/config/config.go
```diff
@@ -127,6 +127,8 @@ type TestEnvironment struct {
 	DeploymentsUnderTest []configsections.Deployment
```
hamadise, line 127: general comments :
The code is this file should also be updated 
```
test-network-function/diagnostic/suite.go
```

pkg/config/configsections/Nodes.go
jun-rh: Two booleans? Isn't it simpler to have just two node list/array?

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,55 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
 	}
 
 	target.DeploymentsUnderTest = append(target.DeploymentsUnderTest, FindTestDeployments(labels, target, namespace)...)
+	target.Nodes = append(target.Nodes, GetNodesList()...)
+}
+
+// GetNodesList Function that return a list of node and what is the type of them.
+func GetNodesList() (nodes []configsections.Nodes) {
```
jun-rh, line 77: The intent is to discover all nodes that have workload running from the CNF under test, not the full list of nodes.

jun-rh: We already capture node name in the pod data structure. I think we just need to process that and come up with a node list/map. No need to create new handler or invoke new commands. In the future, if we need more info about the nodes, we can invoke get node command then

pkg/config/configsections/Nodes.go
hamadise: we had a discussion about this. 
a node can have multiple types (storage, worker..) 
a better solution would be to store all those values in Type slice. But we will have two functions isMaster() and isWorker()
having two maps  is not good idea since some tests may need to loop through all the nodes and having two loops would make things long

jun-rh: Sure, I just don't like two bools as it created invalid combinations.

pkg/config/configsections/Node.go
```diff
@@ -0,0 +1,43 @@
...
+package configsections
+
+// Nodes defines in the cluster. with name of the node and the type of this node master/worker,,,,,
+type Node struct {
+	Name string
+	Type []string
+}
+
+var worker = "worker"
+var master = "master"
```
hamadise, line 26: make them as const in the top and place them at the top of the code.

pkg/config/configsections/Node.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (Node) IsMaster(node Node) bool {
+	for i := range node.Type {
+		if node.Type[i] == master {
+			return true
+		}
+	}
+	return false
+}
+func (Node) IsWorker(node Node) bool {
```
hamadise, line 36: the function should not have node as parameter. 
it should have a signature like ``func (n Node) isWorker() bool{}``

pkg/config/configsections/Node.go
```diff
@@ -0,0 +1,43 @@
...
+	for i := range node.Type {
+		if node.Type[i] == worker {
```
hamadise, line 38: you can get directly the type using ``for _, type := range node.Type {}``
it;s better than getting the index, then getting the type

test-network-function/diagnostic/suite.go
```diff
@@ -185,14 +186,22 @@ func getFirstNode(labelFilter map[string]*string) string {
...
-func getWorkerNodeName() string {
-	const workerNodeLabel = "node-role.kubernetes.io/worker"
-	return getFirstNode(map[string]*string{workerNodeLabel: nil})
+func getWorkerNodeName(env *config.TestEnvironment) string {
+	for _, node := range env.Nodes {
+		if node.IsWorker(node) {
+			return node.Name
+		}
+	}
+	return "null"
```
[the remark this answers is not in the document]
hamadise: same thing for MasterNode name

test-network-function/diagnostic/suite.go
```diff
@@ -185,14 +186,22 @@ func getFirstNode(labelFilter map[string]*string) string {
 	return nodeNames[0]
 }
 
-func getMasterNodeName() string {
-	const masterNodeLabel = "node-role.kubernetes.io/master"
-	return getFirstNode(map[string]*string{masterNodeLabel: nil})
+func getMasterNodeName(env *config.TestEnvironment) string {
+	for _, node := range env.Nodes {
```
hamadise, line 190: in here, you can start from the back of the map. 
We should try to avoid returning the same node in master and worker.

pkg/config/configsections/Node.go
```diff
@@ -0,0 +1,43 @@
...
+var master = "master"
+
+// Nodes defines in the cluster. with name of the node and the type of this node master/worker,,,,,
+type Node struct {
+	Name string
+	Type []string
+}
+
+func (node Node) IsMaster() bool {
+	for _, typ := range node.Type {
```
hamadise, line 29: either use ``t`` or ``type``

pkg/config/configsections/Node.go
```diff
@@ -0,0 +1,43 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package configsections
+
+var worker = "worker"
+var master = "master"
```
hamadise, line 20: better to have them as constant

test-network-function/platform/suite.go
```diff
@@ -51,6 +50,8 @@ import (
 	utils "github.com/test-network-function/test-network-function/pkg/utils"
 )
 
+var worker = "worker"
```
hamadise, line 53: variable not used. should be removed.

hamadise (approved):

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,54 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+func GetNodesList() (nodes map[string]configsections.Node) {
+	nodes = make(map[string]configsections.Node)
+	var nodeNames []string
+	var types []string
+	context := common.GetContext()
+	tester := nodenames.NewNodeNames(common.DefaultTimeout, map[string]*string{"node-role.kubernetes.io/master": nil})
+	test, _ := tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
+	_, err := test.Run()
+	if err != nil {
+		log.Error("Unable to get node list ", ". Error: ", err)
```
greyerof, line 86: Does it make sense to continue if test.Run() failed? Maybe you should return after the log.Error().

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,54 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+			}
+			nodes[nodeNames[i]] = node
+		}
+	}
+	context = common.GetContext()
+	tester = nodenames.NewNodeNames(common.DefaultTimeout, map[string]*string{"node-role.kubernetes.io/worker": nil})
+	test, _ = tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
+	_, err = test.Run()
+	if err != nil {
+		log.Error("Unable to get node list ", ". Error: ", err)
```
greyerof, line 102: Same here, but the nodes variable should be emptied before returning.

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,54 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+	test, _ := tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
+	_, err := test.Run()
+	if err != nil {
+		log.Error("Unable to get node list ", ". Error: ", err)
+	} else {
+		nodeNames = tester.GetNodeNames()
+		for i := range nodeNames {
+			node := configsections.Node{
+				Name: nodeNames[i],
+				Type: append(types, "master"),
```
greyerof, line 92: The "types" variable declared at the begining of this function can be removed, since it's always empty. You can just do:
`Type: []string{"master"}`

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,54 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+				var nod = nodes[nodeNames[i]]
+				nod.Type = append(nod.Type, "worker")
+				nodes[nodeNames[i]] = nod
+
+			} else {
+				node := configsections.Node{
+					Name: nodeNames[i],
+					Type: append(types, "worker"),
+				}
+				nodes[nodeNames[i]] = node
```
greyerof, line 116: You can avoid creating the temporary variable "node" if you do this:
`nodes[nodeNames[i]] = configsections.Node{ Name: nodeNames[i], Type: []string{"worker"})`

pkg/config/autodiscover/autodiscover_targets.go
hamadise: as part of my next task to open debug session will all the nodes, I would need to create "NodeConfig" structure to hold "OC" session. I think I'll get this list of nodes, then add something to indicate if there's any deployment on this node and "OC" session pointer. so it's fine to keep the list of all nodes in here.

Why we want to keep all the nodes?
we do want to test hugepage config and kernel args for all the nodes in the cluster, if CNF scales, we don't know where it will be deployed, hence we need to make sure all nodes are in good state.
Why knowing a deployment is present in node is useful ?
we don't want to drain a node that does not have any deployment.

I would advice to merge this as-is, then work out the details later in my PR.

pkg/config/autodiscover/autodiscover_targets.go
aabughosh: i think there is no need to return here ,,, if we got here it mean that we have master node and we wanto go to the final return that we have in this function

greyerof: You're right. I just noticed that with a return here you can save one indentation because the "else" would not be needed for that part.

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,53 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+		for i := range nodeNames {
+			if _, ok := nodes[nodeNames[i]]; ok {
+				var nod = nodes[nodeNames[i]]
+				nod.Type = append(nod.Type, "worker")
+				nodes[nodeNames[i]] = nod
+
+			} else {
+				nodes[nodeNames[i]] = configsections.Node{
+					Name: nodeNames[i],
+					Type: append(types, "worker"),
```
greyerof, line 114: You can also avoid using the "types" variable here. It's always empty since you're never adding anything to it. In fact, you're creating a new Node here with the "worker" type only, so you can just put:
```
Type: []string{"worker"}
```

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -19,7 +19,11 @@ package autodiscover
 import (
 	log "github.com/sirupsen/logrus"
 	"github.com/test-network-function/test-network-function/pkg/config/configsections"
+	"github.com/test-network-function/test-network-function/pkg/tnf"
+	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/nodenames"
+	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
 	"github.com/test-network-function/test-network-function/pkg/tnf/testcases"
+	"github.com/test-network-function/test-network-function/test-network-function/common"
```
hamadise, line 26: @jun-rh  
I believe the reason why they are using direct 'oc' command in autodiscover package is to avoid cyclic dependency. 
with this change, autodiscover package depends on common package. and it should not be the case

We should move this function : 
```
func GetContext() *interactive.Context {
```
from common package and move it into pkg/interactive/spawner
once this is done, we can stop using "oc" directly without the context

hamadise (changes requested):
we should not create any cyclical dependency. everything under pkg should not depend on code under ./test-network-function

pkg/config/autodiscover/autodiscover_targets.go
jun-rh: Wouldn't it become a problem when you test a tiny cnf on a huge cluster? I think i am ok with doing platform tests only on the nodes running the workloads. We are not doing run time test. We don't need to worry about whether the whole cluster is setup in the same way. We are testing that the workload does not require non standard node config to run.

hamadise: If we have tiny CNF, we will spend more time testing Nodes that will not be used, but I think that's fine since the tests are not a run time program but rather something that should be run couple of times at most.

pkg/config/autodiscover/autodiscover_targets.go
hamadise: I can make this change later in my PR. so don't worry about it.

pkg/config/autodiscover/autodiscover_targets.go
hamadise: In all cases, I think it's good to fetch the list of nodes, then filter it based on our need.

jun-rh: Ok, we can build the full list as long as there is a way to pick the right one from the list by the tests

pkg/config/autodiscover/autodiscover_targets.go
jun-rh: Not sure if it really was the reason. Before I centralized all the spawn local shell calls to common.GetContext(), the contexts were created in many different ways. Some were just a spawn shell call. Some have one or two assertions after. Some suites had a local getContext() function with 3 assertions. If they were to use handler in autodiscover, they probably wouldn't be afraid of repeating a few lines there. Anyway, I am not sure if we want to use gomega in the pkg/tnf/interactive. Right now it's only used in pkg/config and pkg/tnf/test.go

pkg/config/autodiscover/autodiscover_targets.go
jun-rh: @hamadise Are you going to update the tests that use node list? If so, Amal's PR can be merged without that part of update.

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -66,6 +70,51 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
...
+	tester = nodenames.NewNodeNames(common.DefaultTimeout, map[string]*string{"node-role.kubernetes.io/worker": nil})
+	test, _ = tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
+	_, err = test.Run()
+	if err != nil {
+		log.Error("Unable to get node list ", ". Error: ", err)
+	} else {
+		nodeNames = tester.GetNodeNames()
+		for i := range nodeNames {
+			if _, ok := nodes[nodeNames[i]]; ok {
+				var nod = nodes[nodeNames[i]]
```
jun-rh, line 106: nod -> node

test-network-function/platform/suite.go
```diff
@@ -287,66 +286,39 @@ func testSysctlConfigsHelper(podName, podNamespace string) {
...
-		gomega.Expect(nodeNames).NotTo(gomega.BeNil())
-		ginkgo.By("Should not have tainted nodes")
-		defer results.RecordResult(identifiers.TestNonTaintedNodeKernelsIdentifier)
-		if len(nodeNames) == 0 {
-			ginkgo.Skip("Can't test tainted nodes when list of nodes is empty. Please check previous tests.")
-		}
+
 		var taintedNodes []string
-		for _, node := range nodeNames {
+		for _, node := range env.Nodes {
```
jun-rh, line 295: Should be filtered node list.

jun-rh: also, other platform tests that also use node list should be updated.

pkg/config/autodiscover/autodiscover_targets.go
hamadise: Yes. I'm updating it.

pkg/config/autodiscover/autodiscover_targets.go
hamadise: right now, with this change, the autodiscovery code will depend on "common" which in itself depends on gomega. The autodiscovery code should not depend on the "common" code. and we should create our own session instead of getting it from common.

test-network-function/platform/suite.go
jun-rh: Other test requires filtering based on whether the cnf work load is present. Will be done by @hamadise

pkg/config/configsections/Node.go
```diff
@@ -0,0 +1,46 @@
...
+
+package configsections
+
+const worker = "worker"
+const master = "master"
+
+// Node defines in the cluster. with name of the node and the type of this node master/worker,,,,.
+type Node struct {
+	Name string
+	Type []string
```
jun-rh, line 25: This should be plural. Also, why not just call it labels to match the k8s manifest?

hamadise (approved):

pkg/config/configsections/Node.go
aabughosh: done

jun-rh (approved):

PR #374, opened 2021-09-29
Stop error channel monitoring routing prior to scaling in

jun-rh:


greyerof (approved):

test-network-function/lifecycle/suite.go
```diff
@@ -313,7 +326,7 @@ func testPodsRecreation(env *config.TestEnvironment) {
 	})
 }
 
-func closeOcSessions(containers map[configsections.ContainerIdentifier]*config.Container, nodeName string) {
+func closeOcSessionsByNode(containers map[configsections.ContainerIdentifier]*config.Container, nodeName string) {
```
hamadise, line 329: it's better to place ```closeOcSessionsByNode``` and ```closeOcSessionsByDeployment``` since they are doing almost the same thing

hamadise (approved):

edcdavid (approved):
Why not just delete all sessions? We are re-establishing them anyways after rediscovery.

jun-rh:
> Why not just delete all sessions? We are re-establishing them anyways after rediscovery. no need to change it

I want to minimize the chance of shadowing real problems.

test-network-function/lifecycle/suite.go
jun-rh: updated

PR #376, opened 2021-09-29
README update for session timeout requirement

jun-rh:


README.md
```diff
@@ -159,7 +159,8 @@ The image can be pulled using :
 docker pull quay.io/testnetworkfunction/test-network-function
 ```
 ### Cluster requirement
-To run all the tests, OCP cluster should provide enough resources to drain nodes and reschedule pods. If that's not the case, then ``lifecycle-pod-recreation`` test should be skipped.
+* OCP cluster should allow interactive shell sessions to pods/containers to stay alive when being idle for more than a few minutes. If it is not the case, consult the maintainer of the cluster infrastructure on how it can be enabled. It could also be interfered by any firewall/load balancer that sits between the cluster API server and where the test runner is.
```
edcdavid, line 162: It could also be interfered by any firewall/load balancer that sits between the cluster API server and where the test runner is.
->
Even if OCP cluster is configured with a longer idle timeout, a firewall/load balancer configured between the cluster API server and the test runner could still introduce a short idle timeout. Consult with the maintainer of the load balancer on how to increase it.

edcdavid (approved):

PR #377, opened 2021-09-29
Adding Support for exit code in goexpect/commands and minor typo fixes to catalog

edcdavid:


pkg/tnf/testcases/base.go
```diff
@@ -176,16 +176,18 @@ const (
...
-	EmptyNullFalse: `^\b(null|false)\b$|^$`,
-	NullFalse:      `^\b(null|false)\b$`,
-	True:           `^\b(true)\b$`,
-	Null:           `^\b(null)\b$`,
-	Zero:           `0`,
-	NonZeroNumber:  `^(0*([1-9]\d*)|null)$`,
-	Error:          "error",
-	Digit:          `\d`,
+	// Removing all empty cases as they are not used and will break the parsing
+	// AllowEmpty:     `(.*?)`,
```
jun-rh, line 180: Just want to make sure, we do still support this case? this is just to get around unit test when terminal simulation is off?

jun-rh: Never mind. I thought this was for unit test. Now I see it's for those special combo tests.

pkg/tnf/reel/reel.go
```diff
@@ -254,15 +279,31 @@ func (r *Reel) wrapTestCommand(cmd string) string {
...
+	return wrappedCommand
 }
 
 // stripEmulatedPromptFromOutput will elide the emulated terminal prompt from the test output.
-func (r *Reel) stripEmulatedPromptFromOutput(output string) string {
-	if !r.disableTerminalPromptEmulation {
-		return strings.TrimRight(strings.TrimRight(output, endOfTestSentinelCutset), "\n")
+func (r *Reel) stripEmulatedPromptFromOutput(output string) (data string, status int) {
+	parsed := strings.Split(output, EndOfTestSentinel)
+	if !r.disableTerminalPromptEmulation && len(parsed) > 1 {
```
jun-rh, line 290: = 2 instead? that's the only case supported by this branch, isn't it?

jun-rh: also, can we still support empty data, i.e. len =1? just use parsed[0] for status parsing and set data to ""?

pkg/tnf/reel/reel.go
```diff
@@ -123,10 +137,15 @@ func (r *Reel) batchExpectations(expectations []string, batcher []expect.Batcher
...
 	var cases []expect.Caser
+	// expectations created from test case matches
 	for _, expectation := range expectations {
 		thisCase := r.generateCase(expectation, firstMatch)
 		cases = append(cases, thisCase)
 	}
+	// extra test case to match when commands do not return anything but exit without error. This expectation makes
+	// sure that any command exiting successfully will be processed without timeout.
+	thisCase := r.generateCase("", firstMatch)
+	cases = append(cases, thisCase)
```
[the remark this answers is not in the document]
edcdavid: I retested test cases that return "" and they work now. must be one of the later fix. The additional testcase is still needed if the command returns nothing unexpectedly. this is the last expectation if every thing else fails or timeouts to be able to read the status.

pkg/tnf/reel/reel.go
edcdavid: Fixed ==2
The case with empty data is covered also by ==2. parsed[0] will be "" in this case

jun-rh: cool, i thought this was the reason why empty patterns were no longer supported. I see you have added them back.

pkg/tnf/reel/reel.go
```diff
@@ -199,7 +221,8 @@ func (r *Reel) Step(step *Step, handler Handler) error {
 				} else {
 					before = ""
 				}
-				step = handler.ReelMatch(r.stripEmulatedRegularExpression(firstMatch), before, match)
+				amatch := r.stripEmulatedRegularExpression(firstMatch)
```
jun-rh, line 224: I got a bit confused by firstMatch, amatch, match. Took me a while to realize firstMatch and amatch are regex patterns. Maybe name them differently to reflect that?

test-network-function/diagnostic/suite.go
```diff
@@ -216,8 +216,8 @@ func testOcpVersion() {
 	test, err := tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
 	gomega.Expect(err).To(gomega.BeNil())
 	testResult, err := test.Run()
```
jun-rh, line 218: there is a RunAndValidate() that does what these 3 lines do

pkg/tnf/reel/reel.go
edcdavid: fixed

test-network-function/diagnostic/suite.go
edcdavid: ok sure. 
Yes, I believe the logging is needed. It is not captured otherwise.

PR #379, opened 2021-10-01
Get Release version from Git tag

edcdavid:


Makefile
```diff
@@ -96,11 +98,11 @@ build-catalog-md:
 
 # build the CNF test binary
 build-cnf-tests:
-	PATH=${PATH}:${GOBIN} ginkgo build -ldflags "-X github.com/test-network-function/test-network-function/test-network-function.GitCommit=${GIT_COMMIT}" ./test-network-function 
+	PATH=${PATH}:${GOBIN} ginkgo build -ldflags "-X github.com/test-network-function/test-network-function/test-network-function.GitCommit=${GIT_COMMIT} -X github.com/test-network-function/test-network-function/test-network-function.GitRelease=${GIT_RELEASE} -X github.com/test-network-function/test-network-function/test-network-function.GitPreviousRelease=${GIT_PREVIOUS_RELEASE}" ./test-network-function
```
hamadise, line 101: we not just use "commit=" and "version=" ?
why we need previous release as well ?

Makefile
```diff
@@ -96,11 +98,11 @@ build-catalog-md:
...
+	PATH=${PATH}:${GOBIN} ginkgo build -ldflags "-X github.com/test-network-function/test-network-function/test-network-function.GitCommit=${GIT_COMMIT} -X github.com/test-network-function/test-network-function/test-network-function.GitRelease=${GIT_RELEASE} -X github.com/test-network-function/test-network-function/test-network-function.GitPreviousRelease=${GIT_PREVIOUS_RELEASE}" ./test-network-function 
 	make build-catalog-md
 
 build-cnf-tests-debug:
-	PATH=${PATH}:${GOBIN} ginkgo build -gcflags "all=-N -l" -ldflags "-X github.com/test-network-function/test-network-function/test-network-function.GitCommit=${GIT_COMMIT} -extldflags '-z relro -z now'" ./test-network-function
+	PATH=${PATH}:${GOBIN} ginkgo build -gcflags "all=-N -l" -ldflags "-X github.com/test-network-function/test-network-function/test-network-function.GitCommit=${GIT_COMMIT} -X github.com/test-network-function/test-network-function/test-network-function.GitCommit=${GIT_RELEASE} -X github.com/test-network-function/test-network-function/test-network-function.GitPreviousRelease=${GIT_PREVIOUS_RELEASE} -extldflags '-z relro -z now'" ./test-network-function
```
hamadise, line 105: ditto

test-network-function/suite_test.go
```diff
@@ -115,7 +123,16 @@ func TestTest(t *testing.T) {
 	gomega.RegisterFailHandler(ginkgo.Fail)
 	common.SetLogFormat()
 	common.SetLogLevel()
-	log.Info("Version: ", programVersion, " ( ", GitCommit, " )")
+
+	// Display the latest previously released build in case this build is not released
+	// Otherwise display the build version
+	if GitRelease == "" {
```
hamadise, line 129: I don't see where "GitRelease" is being initialized, it will always be ``==""``

Makefile
edcdavid: The previous version is used to give the last release of a given unrelease build. For instance, currently it will say unreleased, post 3.0.0.

test-network-function/suite_test.go
edcdavid: GitRelease is initialized during the builing process, it is a static variable.

Makefile
hamadise: I think my comment was not clear. 
what you're passing after -X is key=value; the key is very long. we need only 
``` -X commit = ${GIT_COMMIT} Release = ${GIT_RELEASE} -X PreviousRelease = ${}.. ``

edcdavid: This is the name of the static variable object in the executable, not sure how to shorten it. I tried the short but ended up needing a obj decoder to find out the working name. See:
[deliedit@cnfcert-local cnftest]$ go tool nm test-network-function/test-network-function.test  | grep -i gitCommit
  f05540 D github.com/test-network-function/test-network-function/test-network-function.GitCommit
  b450a0 R github.com/test-network-function/test-network-function/test-network-function.GitCommit.str

PR #380, opened 2021-10-06
CRD autodiscovery using spec.group.

greyerof:
Added targetCrds yaml entry for CRDs autodiscovery. Currently, they can
only be discovered using the group (spec.group) tag, but it can be
extended to use others like "name" or just labels as with pods/cuts.

It just creates a list with the names of the CRDs that belong to the
configured groups.


pkg/config/configsections/crd.go
```diff
@@ -0,0 +1,24 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+package configsections
+
+// Crd defines a CustomResourceDefinition config filter.
+type Crd struct {
```
jun-rh, line 20: does this really need to be a structure? even if we add other ways of matching CRDs, it would be a separate list, right? e.g. 
```
targetCrds:
  groups:
    - a
    - b
  labels:
    - name: lname
      value: lvalue
```

PR #382, opened 2021-10-06
CNFCERT-148:Update test suite to support statefulset and stateful app…

aabughosh:
…lications


jun-rh (changes requested):
Update identifiers.go / catalog

pkg/tnf/handlers/owners/owners.go
```diff
@@ -76,7 +76,8 @@ func (ow *Owners) ReelFirst() *reel.Step {
 
 // ReelMatch ensures that list of nodes is not empty and stores the names as []string
 func (ow *Owners) ReelMatch(_, _, match string) *reel.Step {
-	if strings.Contains(match, "ReplicaSet") && !strings.Contains(match, "DaemonSet") {
+	if (strings.Contains(match, "StatefulSet") || strings.Contains(match, "ReplicaSet")) &&
```
hamadise, line 79: it's better to define those as constant in the top of the file

pkg/tnf/handlers/owners/owners.go
```diff
@@ -27,6 +27,12 @@ import (
 
 const (
 	owRegex = "(?s)OWNERKIND\n.+"
+	// StatefulSet variable
```
hamadise, line 30: using Uppercase in golang, means you are exporting those variables outside of this module. 
it's better to minimize the scope of any variable until it's needed.

use lowercase in all the constants

hamadise (approved):

PR #390, opened 2021-10-08
Add yamllint action

sebrandon1:
https://github.com/marketplace/actions/yaml-lint

Adding a GitHub Action that performs linting of YAML files.


sebrandon1:
Looks like my GitHub action is working, however it's actually doing it's job and the lint is failing on our YAMLs.  😄   Will attempt to fix...

PR #391, opened 2021-10-11
CRD status handler using generic json template.

greyerof:
This is just the json spec to test whether a CRD (by name) has a defined status specification. For the actual TC, [autodiscovery PR](https://github.com/test-network-function/test-network-function/pull/380) needs to be merged first. Then, the TC can be implemented in the observability TS iterating through all the CRD names using this handler.


jun-rh:
Will there be an update to the local test infra? Right now, the test must be getting skipped in the workflow.

PR #398, opened 2021-10-18
Integrate best practice reference into test catalog

shimritproj:
Find the best practice reference for test.


sebrandon1:
Just a general question, would it be better to have the strings such as `https://connect.redhat.com/sites/default/files/2021-03/Cloud%20Native%20Network%20Function%20Requirements.pdf (Section 6.2 )` be a variable so if any potential changes are needed its a single line versus many?

shimritproj:
Updated for every test we have a reference: link + section. @jun-rh @sebrandon1

PR #400, opened 2021-10-18
Fix URL for ping test

sebrandon1:
This is supposed to point [here](https://github.com/test-network-function/test-network-function#issue-161-some-containers-under-test-do-not-contain-ping-or-ip-binary-utilities).


sebrandon1:
I have no idea what's going on with the `golangci-lint` curl in the checks, but it seems to work locally.

PR #408, opened 2021-10-25
Adding support for more failure scenarios and logs in the lifecycle suite

edcdavid:
Added a check to make sure there are available schedulable node before draining a existing node
Added more logs and increased scaling timeout


edcdavid:
additional logging will be addressed in another PR. Issues already fixed by other PRs.

PR #424, opened 2021-10-29
Fix for duplicate dangling sessions un-expectedly closing

edcdavid:


pkg/config/config.go
```diff
@@ -232,6 +232,11 @@ func (env *TestEnvironment) doAutodiscover() {
 		env.ContainersToExcludeFromConnectivityTests[cid] = ""
 	}
 
+	// Delete all remaining sessions before re-creating them
```
jun-rh, line 235: Move it up to somewhere around line 206?

jun-rh (approved):

hamadise (approved):

pkg/config/config.go
edcdavid: fixed

PR #428, opened 2021-11-02
Consolidate all cmds into one tnf cmd

aabughosh:


cmd/tnf/addclaim/addclaim.go
```diff
@@ -0,0 +1,113 @@
+package addclaim
+
```
hamadise, line 2: I think the package can be claim instead of addclaim. We can add more functions to this package such as update/modify the claim, so it's better to keep it generic

Makefile
```diff
@@ -85,20 +81,20 @@ test: mocks
 	go test -coverprofile=cover.out `go list ./... | grep -v "github.com/test-network-function/test-network-function/test-network-function" | grep -v mock`
 
 # build the binary that can be used to run JSON-defined tests.
-build-jsontest-cli:
-	go build -o jsontest-cli -v cmd/generic/main.go
+#build-jsontest-cli:
```
jun-rh, line 84: No need to keep these as commented, just remove them

cmd/tnf/main.go
```diff
@@ -1,151 +1,40 @@
...
-		return err
-	}
-	w.Flush()
-
-	return nil
-}
-
 func main() {
+	claimAddFile := addclaim.NewCommand()
+	rootCmd.AddCommand(claimAddFile)
```
jun-rh, line 28: no need for local variables. it's easier to read if you just do:

```suggestion
	rootCmd.AddCommand(claim.NewCommand())
```

jun-rh (changes requested):
Also, update developing.md

jun-rh (approved):

hamadise (approved):

PR #436, opened 2021-11-02
exclude default pod IP from multus ip list

edcdavid:
If no multus IPs are found, the multus test is skipped


hamadise:
@edcdavid  can you rebase this ? I believe Jun did similar changes to yours. maybe it's not even needed

hamadise (approved):

jun-rh (approved):

jun-rh:
> @edcdavid can you rebase this ? I believe Jun did similar changes to yours. maybe it's not even needed

I did the normal one only as I knew David was working on the multus one. And I thought he was gonna make it not use partner pod. Otherwise I would have done both.

PR #438, opened 2021-11-03
use number of nodes with expected debug pod to verify if daemonset is…

hamadise:
… ready

### DIFF
--- a/pkg/config/autodiscover/autodiscover_debug.go
+++ b/pkg/config/autodiscover/autodiscover_debug.go
@@ -77,15 +77,15 @@ func DeleteDebugLabel(nodeName string) {
 
 // CheckDebugDaemonset checks if the debug pods are deployed properly
 // the function will try DefaultTimeout/time.Second times
-func CheckDebugDaemonset() {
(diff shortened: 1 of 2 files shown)


pkg/config/autodiscover/autodiscover_debug.go
```diff
@@ -98,9 +98,11 @@ func checkDebugPodsReadiness() bool {
 		return false
 	}
 	dsStatus := tester.GetStatus()
-	if dsStatus.Desired == dsStatus.Current &&
+	if dsStatus.Desired == expectedDebugPods &&
+		dsStatus.Desired == dsStatus.Current &&
 		dsStatus.Available == dsStatus.Ready &&
 		dsStatus.Ready == dsStatus.Desired &&
+
 		dsStatus.Ready != 0 &&
```
jun-rh, line 106: this check is no longer needed?

jun-rh (approved):

pkg/config/autodiscover/autodiscover_debug.go
hamadise: fixed

greyerof (approved):

PR #448, opened 2021-11-04
Fix access-control HOST_PORT_CHECK tc.

greyerof:
When containers have ports declared, the previous jq query failed
to get them with this error:
jq: error (at <stdin>:1330): Cannot index array with string "hostPort"

That's because of the query is not accessing correctly to the ports
array, so the TC would fail either its ports are hostPort or
containerPort ones.


greyerof:
This task is already being addressed here: https://github.com/test-network-function/test-network-function/pull/444

PR #451, opened 2021-11-04
Fix error handling in platform tests

jun-rh:


pkg/tnf/test.go
```diff
@@ -135,15 +135,23 @@ func (t *Test) RunAndValidateWithFailureCallback(cb func()) {
 	gomega.Expect(err).To(gomega.BeNil())
 }
 
-// RunWithFailureCallback runs the test, invokes the cb on failure
+// RunWithCallbacks runs the test, invokes the cb on failure/error/success
 // This is useful when the testcase needs to continue whether this test result is success or not
-func (t *Test) RunWithFailureCallback(cb func()) {
+func (t *Test) RunWithCallbacks(scb, fcb func(), ecb func(error)) {
```
hamadise, line 140: scb ? ecb ? what do they stand for ?

hamadise: success call back ? error call back ?

pkg/tnf/test.go
```diff
@@ -135,15 +135,23 @@ func (t *Test) RunAndValidateWithFailureCallback(cb func()) {
...
+			scb()
+		}
+	case FAILURE:
+		if fcb != nil {
+			fcb()
+		}
+	case ERROR:
+		if ecb != nil {
+			ecb(err)
+		}
```
hamadise, line 154: I think generic error printing if the cb is nil would be useful

pkg/tnf/test.go
jun-rh: yes, i can expand the name if it's not clear

pkg/tnf/test.go
jun-rh: generic case should call RunAndValidate() instead

pkg/tnf/test.go
jun-rh: updated

hamadise (approved):

greyerof (approved):

PR #455, opened 2021-11-05
add multinamespaces support

hamadise:


pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -46,37 +46,39 @@ var (
...
 // using labels and annotations, and add them to the `configsections.TestTarget` passed in.
-func FindTestTarget(labels []configsections.Label, target *configsections.TestTarget, namespace string) {
-	// find pods by label
-	for _, l := range labels {
-		pods, err := GetPodsByLabel(l, namespace)
-		if err == nil {
-			for i := range pods.Items {
-				target.PodsUnderTest = append(target.PodsUnderTest, buildPodUnderTest(pods.Items[i]))
-				target.ContainerConfigList = append(target.ContainerConfigList, buildContainersFromPodResource(pods.Items[i])...)
+func FindTestTarget(labels []configsections.Label, target *configsections.TestTarget, namespace []string) {
```
jun-rh, line 49: namespaces

jun-rh:
Also update access-control testNamespace

test-network-function/lifecycle/suite.go
```diff
@@ -313,16 +319,16 @@ func testPodsRecreation(env *config.TestEnvironment) {
...
-
-			waitForAllDeploymentsReady(env.NameSpaceUnderTest, scalingTimeout, scalingPollingPeriod)
-
-			// verify deployments are ready again
-			_, notReadyDeployments = getDeployments(env.NameSpaceUnderTest)
-			if len(notReadyDeployments) != 0 {
-				uncordonNode(n.Name)
-				ginkgo.Fail(fmt.Sprintf("did not create replicas when node %s is drained", n.Name))
+			for _, ns := range env.NameSpacesUnderTest {
+				waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
```
jun-rh, line 323: I wonder if we should check all namespaces. Or at least the ocp namespaces when draining master node.

sebrandon1 (commented):
Some comments!

test-network-function/lifecycle/suite.go
```diff
@@ -293,14 +293,20 @@ func testPodsRecreation(env *config.TestEnvironment) {
...
-		ginkgo.By(fmt.Sprintf("test deployment in namespace %s", env.NameSpaceUnderTest))
-		deployments, notReadyDeployments = getDeployments(env.NameSpaceUnderTest)
-		if len(deployments) == 0 {
-			return
+		ginkgo.By(fmt.Sprintf("test deployment in namespace %s", env.NameSpacesUnderTest))
+		for _, ns := range env.NameSpacesUnderTest {
+			var dps dp.DeploymentMap
+			dps, notReadyDeployments = getDeployments(ns)
+			for dpKey, dp := range dps {
+				deployments[dpKey] = dp
```
javierpena, line 301: Based on the failure in the unit test, we need to initialize deployments in its declaration (line 290 in the file). This was not an issue previously, because we got it from getDeployments().

test-network-function/lifecycle/suite.go
hamadise: can we limit the drain to worker node only ? In real deployment master node should not have any deployment under test on it

test-network-function/lifecycle/suite.go
hamadise: fixed

test-network-function/lifecycle/suite.go
```diff
@@ -313,16 +319,16 @@ func testPodsRecreation(env *config.TestEnvironment) {
...
-			_, notReadyDeployments = getDeployments(env.NameSpaceUnderTest)
-			if len(notReadyDeployments) != 0 {
-				uncordonNode(n.Name)
-				ginkgo.Fail(fmt.Sprintf("did not create replicas when node %s is drained", n.Name))
+			for _, ns := range env.NameSpacesUnderTest {
+				waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
+
+				// verify deployments are ready again
+				_, notReadyDeployments = getDeployments(ns)
+				if len(notReadyDeployments) != 0 {
```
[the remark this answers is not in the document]
hamadise: fixed

pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -46,37 +46,39 @@ var (
...
-	if err == nil {
-		for i := range csvs.Items {
-			target.Operators = append(target.Operators, buildOperatorFromCSVResource(&csvs.Items[i]))
+		csvs, err := GetCSVsByLabel(operatorLabelName, anyLabelValue, ns)
+		if err == nil {
+			for i := range csvs.Items {
+				target.Operators = append(target.Operators, buildOperatorFromCSVResource(&csvs.Items[i]))
+			}
+		} else {
+			log.Warnf("an error (%s) occurred when looking for operaters by label", err)
```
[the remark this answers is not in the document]
hamadise: fixed

test-network-function/lifecycle/suite.go
jun-rh: That was the next thought I had too after leaving the comment :)
Then SNO and 3 node clusters came to my mind ...
We would definitely skip intrusive tests on SNO. 3 node cluster though seems to be a real thing for some edge scenarios.

jun-rh: If the current implementation works reliably on our 3 node cluster, mark this one as resolved

pkg/config/config.go
```diff
@@ -247,12 +248,12 @@ func (env *TestEnvironment) ResetOc() {
 
 func (env *TestEnvironment) doAutodiscover() {
 	log.Debug("start auto discovery")
-	if len(env.Config.TargetNameSpaces) != 1 {
-		log.Fatal("a single namespace should be specified in config file")
+	for _, ns := range env.Config.TargetNameSpaces {
+		env.NameSpacesUnderTest = append(env.NameSpacesUnderTest, ns.Name)
```
edcdavid, line 252: add 	env.NameSpacesUnderTest=nil in reset(), otherwise autodiscovery will keep adding the same namespace each time this is called. See below:
func (env *TestEnvironment) reset() {
	log.Debug("clean up environment Test structure")
	env.ResetOc()
	env.Config.Partner = configsections.TestPartner{}
	env.Config.TestTarget = configsections.TestTarget{}
	env.TestOrchestrator = nil
	env.NameSpacesUnderTest=nil
This should fix the smoke test

edcdavid (approved):

pkg/config/config.go
hamadise: fixed

jun-rh (approved):

PR #467, opened 2021-11-12
Escape the new line char in the utils.ExecuteCommand function.

greyerof:
Apart from the double quote, the new line char also needs to be escaped
so the command doesn't break the json string value for the generic
Command handler.

Example: "oc get ... -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}'
         The {{"\n"}} needs to be replaced to {{\\"\\\n\\"}}


sebrandon1 (approved):

pkg/utils/utils.go
```diff
@@ -67,10 +67,13 @@ func ExecuteCommand(command string, timeout time.Duration, context *interactive.
 	log.Debugf("Executing command: %s", command)
 
 	values := make(map[string]interface{})
-	// Escapes the double quote char to make a valid json string.
-	values["COMMAND"] = strings.ReplaceAll(command, "\"", "\\\"")
+	// Escapes the double quote and new line chars to make a valid json string for the command to be executed by the handler.
+	values["COMMAND"] = strings.ReplaceAll(command, "\"", `\"`)
+	values["COMMAND"] = strings.ReplaceAll(values["COMMAND"].(string), "\n", `\\n`)
```
edcdavid, line 72: Should we also escape the other characters at https://www.freeformatter.com/json-escape.html ?

sebrandon1 (approved):

pkg/utils/utils.go
greyerof: Added function to escape chars through marshalling.

jun-rh (approved):

sebrandon1 (approved):

edcdavid (approved):

PR #468, opened 2021-11-12
Namespaces TC extended with CRs namespace check.

greyerof:
This PR makes the following modifications to the namespaces TC:
1. Check the configured namespaces (env.NamespacesUnderTest) to not have invalid prefixes.
2. Iterate through discovered pods and CRs to check whether they belong to those namespaces.
2.1. pods' namespaces check: Fail if len(env.NonCnfPods) > 0. Show all the NonCnfPods for debugging. The env.NonCnfPods is built during autodiscovery with all the pods found with the discovery labels that don't belong to the configured namespaces.
2.2. crs' namespaces check: find all CRs belonging to the discovered CRDs and check that their namespaces belong to the configured list.


greyerof:
This PR needs [PR467](https://github.com/test-network-function/test-network-function/pull/467) to be merged first in order to pass the smoke tests.

greyerof:
As per istio docs (https://istio.io/latest/docs/setup/getting-started/), it will create the namespace "istio-system" when installed. Not sure if more namespaces will be created when configuring/upgrading it, but seems reasonable to me to filter using "istio-" as prefix for invalid namespaces.
I still haven't found the namespace that aspen mesh uses. They also seem to use istio-system, but I'm still investigating this.

test-network-function/accesscontrol/suite.go
```diff
@@ -31,11 +31,34 @@ import (
...
 	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
 	"github.com/test-network-function/test-network-function/pkg/tnf/testcases"
+	"github.com/test-network-function/test-network-function/pkg/utils"
 	"github.com/test-network-function/test-network-function/test-network-function/common"
 	"github.com/test-network-function/test-network-function/test-network-function/identifiers"
 	"github.com/test-network-function/test-network-function/test-network-function/results"
 )
 
+const (
+	ocGetCrKindFormat = "oc get crd %s -o jsonpath='{.spec.names.kind}'"
```
jun-rh, line 41: use plural or singular. i happened to notice that network-attachment-definitions.k8s.cni.cncf.io CRD has a different name as kind, which can not be used as resource type.

test-network-function/accesscontrol/suite.go
```diff
@@ -31,11 +31,34 @@ import (
...
+
+	// ocGetCrNamespaceFormat is the "oc get" format string to get the namespaced-only resources created for a given CRD.
+	ocGetCrNamespaceFormat = "oc get %s -A -o go-template='{{range .items}}{{if .metadata.namespace}}{{.metadata.name}},{{.metadata.namespace}}{{\"\n\"}}{{end}}{{end}}'"
+)
+
+var (
+	invalidCrNamespacePrefixes = []string{
+		"istio-",
```
jun-rh, line 49: default and openshift should be checked here too.

jun-rh (changes requested):
does the catalog need an update too?

greyerof:
After talking to @jun-rh , this TC will be changed to do the following:

Pre-condition (different PR, by @hamadise):
The current PuTs autodiscovery query needs to be modified to make a label-only filter search. All these label-based discovered pods will be filtered using the env.NamespacesUnderTest list to build the final PuT list. Discarded pods should be appeneded to a new env.NonCnfPods (better name needed?) list.

Changes to this PR:
1. Check the configured namespaces (env.NamespacesUnderTest) to not have invalid prefixes.
2. Iterate through discovered pods and CRs to check whether they belong to those namespaces.
	2.1. pods' namespaces check (see pre-condition): Fail if len(env.NonCnfPods) > 0. Show all the NonCnfPods for debugging.
	2.2. crs' namespaces check: find all CRs belonging to the discovered CRDs and check that their namespaces belong to the configured list.
3. Update README.md

So, this PR will depend on the pre-condition PR and won't be merged until that one has been merged.

Extra (new PRs):
- Modify the CRD autodiscovery to get both the target CRD name and its pluralName to avoid doing the extra pluralName query in this TC.
- Break this TC into 3 different TCs.
  + Configured namespaces validity check.
  + Pods' namespaces check.
  + CRs' namespaces check.

jun-rh (approved):

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
+	cmdOut := utils.ExecuteCommand(getCrNamespaceCommand, common.DefaultTimeout, common.GetContext(), func() {
+		tcClaimLogPrintf("CRD %s: Failed to get CRs (kind=%s)", crdName, crdKind)
+	})
+	// No CRs created for this CRD yet.
+	if cmdOut == "" {
+		return map[string]string{}
+	}
+	lines := strings.Split(cmdOut, "\n")
+
+	crNamespaces := map[string]string{}
```
hamadise, line 171: move this to line 165, returning it as an empty list in line 167 or fill it with values in there's any

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
+	for _, line := range lines {
+		if line == "" {
+			// This is to filter the last line, which is always empty.
+			continue
+		}
+		lineFields := strings.Split(line, ",")
+		crNamespaces[lineFields[0]] = lineFields[1]
```
hamadise, line 178: is it safe to assume lineFields will always have length ==2 ? I would prefer having a check for safety

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
 }
 
+func getCrsNamespaces(crdName, crdKind string) map[string]string {
+	gomega.Expect(crdKind).NotTo(gomega.BeEmpty())
+	getCrNamespaceCommand := fmt.Sprintf(ocGetCrNamespaceFormat, crdKind)
+	cmdOut := utils.ExecuteCommand(getCrNamespaceCommand, common.DefaultTimeout, common.GetContext(), func() {
+		tcClaimLogPrintf("CRD %s: Failed to get CRs (kind=%s)", crdName, crdKind)
+	})
+	// No CRs created for this CRD yet.
+	if cmdOut == "" {
```
hamadise, line 166: looking at the whole code, this check is actually not needed. you can remove it and the same result will be returned

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
+		})
+		crNamespaces := getCrsNamespaces(crdName, crdPluralName)
+
+		ginkgo.By(fmt.Sprintf("CRD %s has %d CRs (plural name: %s).", crdName, len(crNamespaces), crdPluralName))
+		for crName, namespace := range crNamespaces {
+			ginkgo.By(fmt.Sprintf("Checking CR %s - Namespace %s", crName, namespace))
+			found := false
+			for _, configNamespace := range configNamespaces {
+				if namespace == configNamespace {
+					found = true
```
hamadise, line 199: add break here to save some time

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
+				ginkgo.By(fmt.Sprintf("Checking namespace %s", namespace))
+				for _, invalidPrefix := range invalidNamespacePrefixes {
+					if strings.HasPrefix(namespace, invalidPrefix) {
+						tcClaimLogPrintf("Namespace %s has invalid prefix %s", namespace, invalidPrefix)
+						failedNamespaces = append(failedNamespaces, namespace)
+					}
+				}
+			}
+			failedNamespacesNum := len(failedNamespaces)
+			if failedNamespacesNum > 0 {
```
hamadise, line 232: failedNamespacesNum is used once. it would be better to use len(failedNamespaces) directly

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
+				ginkgo.Fail(fmt.Sprintf("Found %d namespaces with an invalid prefix.", failedNamespacesNum))
+			}
+
+			ginkgo.By(fmt.Sprintf("CNF pods' should belong to any of the configured namespaces: %v", env.NameSpacesUnderTest))
+			nonValidPodsNum := len(env.Config.NonValidPods)
```
hamadise, line 237: ditto

test-network-function/accesscontrol/suite.go
```diff
@@ -130,16 +156,103 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
+			if nonValidPodsNum > 0 {
+				for _, invalidPod := range env.Config.NonValidPods {
+					tcClaimLogPrintf("Pod %s has invalid namespace %s", invalidPod.Name, invalidPod.Namespace)
+				}
+
+				ginkgo.Fail(fmt.Sprintf("Found %d pods under test belonging to invalid namespaces.", nonValidPodsNum))
```
hamadise, line 243: I'm afraid this can happen in our partners lab. 
If partner is using multi namespace for CNF, then testing a single namespace only. There's a good chance the same labels are shared across namespaces and we will fail this test for no valid reason.

jun-rh: I don't think labels would be shared across namespaces. It's possible though that they put all the labels in one config and only update the namespace list. Not sure if we really have to allow that. I don't think it's too much to ask them to update the label list as well.

test-network-function/accesscontrol/suite.go
greyerof: I've used a custom go-template for the oc get command so the output should always be "crName,namespace" per line, but I agree a safety check is a always a good idea.

hamadise (approved):

PR #470, opened 2021-11-12
Fails Ginko test with non zero exit code

edcdavid:


pkg/tnf/reel/reel.go
```diff
@@ -214,6 +215,10 @@ func (r *Reel) Step(step *Step, handler Handler) error {
 				}
 				match, matchStatus := r.stripEmulatedPromptFromOutput(result.Match[0])
 				log.Debugf("command status: output=%s, match=%s, outputStatus=%d, matchStatus=%d", output, match, outputStatus, matchStatus)
+				// If the exit code is not zero, then fail the test
+				if outputStatus != 0 {
+					ginkgo.Fail(fmt.Sprintf("Failed because of non-zero exit status = %d", outputStatus))
+				}
```
hamadise, line 221: is this intended to be permanent fix or temporary ? if temporary, then that's fine.
Otherwise, this function should return an error, then deal with the error at upper layer.
we should not be adding ginkgo dependency here.

edcdavid: fixed

pkg/tnf/reel/reel.go
```diff
@@ -214,6 +214,10 @@ func (r *Reel) Step(step *Step, handler Handler) error {
 				}
 				match, matchStatus := r.stripEmulatedPromptFromOutput(result.Match[0])
 				log.Debugf("command status: output=%s, match=%s, outputStatus=%d, matchStatus=%d", output, match, outputStatus, matchStatus)
+				// If the exit code is not zero, then fail the test
+				if outputStatus != 0 {
```
jun-rh, line 218: We already check and set r.Err. I think what's missing is the check on r.Err at the end, outside the loop. Looks like a miss from the original design. The original implementation only checks at the beginning of the loop, so the err from the last step gets ignored.

jun-rh: also, line 76-78 in test.go can be removed after fixing this?

jun-rh (changes requested):

pkg/tnf/reel/reel.go
edcdavid: fixed

hamadise (approved):

pkg/tnf/reel/reel.go
```diff
@@ -210,10 +210,12 @@ func (r *Reel) Step(step *Step, handler Handler) error {
 
 				output, outputStatus := r.stripEmulatedPromptFromOutput(result.Output)
 				if outputStatus != 0 {
-					r.Err = fmt.Errorf("error executing command %d: ", outputStatus)
+					r.Err = fmt.Errorf("error executing command exit code:%d", outputStatus)
+					return r.Err
```
jun-rh, line 214: I think the right thing to do is continue here and return r.Err outside the loop. There seems to be other cases where r.Err get set. It's wrong to ignore it for the last step, which in our case, the only step most of the time.

edcdavid: fixed

jun-rh (approved):

PR #472, opened 2021-11-12
Add GOPROXY support

sebrandon1:
Related to: https://issues.redhat.com/browse/CNFCERT-192

Taking advantage of the `proxy.golang.org` to help us with cached dependencies.

Interesting reads:
- https://jfrog.com/blog/why-goproxy-matters-and-which-to-pick/
- https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy/

Marking WIP for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@
 # development. This includes running configured linters and executing unit tests
 
 GO_PACKAGES=$(shell go list ./... | grep -v vendor)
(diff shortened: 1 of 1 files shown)


sebrandon1:
Just kidding, this is already the default value after reading more about the default Go module.  Apparently the "free" GOPROXY only caches for a short amount of time anyways if there is a missing dependency so it might not be what we want long term.

PR #473, opened 2021-11-12
update readme

hamadise:


README.md
```diff
@@ -55,34 +60,6 @@ targetCrdFilters:
...
 
 ### testTarget
-#### podsUnderTest / containersUnderTest
-This section is usually not required if labels defined in the section above cover all resources that should be tested. If label based discovery is not sufficient, this section can be manually populated as shown in the commented part of the [sample config](test-network-function/tnf_config.yml). However, intrusive tests need to be skipped ([see here](#disable-intrusive-tests)) for a reliable test result. The pods and containers explicitly configured here are added to the target pod/container lists populated through label matching.
-
-For both configured and discovered pods/containers, the autodiscovery mechanism will attempt to identify the default network device and all the IP addresses of the pods it
-needs for network connectivity tests, though that information can be explicitly set using annotations if needed. For Pod IPs:
-
-* The annotation `test-network-function.com/multusips` is the highest priority, and must contain a JSON-encoded list of
```
jun-rh, line 64: these annotations are still valid i think

README.md
```diff
@@ -26,9 +26,14 @@ The Test Network Function support autodiscovery using labels and annotations. Th
...
+``` shell script
+targetNameSpaces:
+  - name: firstnamespace
+  - name: secondnamespace 
+```
 ### targetPodLabels
-The goal of this section is to specify the label to be used to identify the CNF resources under test. It's highly recommended that the labels should be defined in pod definition rather than added after pod is created, as labels added later on will be lost in case the pod gets rescheduled. In case of pods defined as part a deployment, it's best to use the same label as the one defined in the `spec.selector.matchLabels` section of the deployment yaml. The prefix field can be used to avoid naming collision with other labels.
+The goal of this section is to specify the labels to be used to identify the CNF resources under test. It's highly recommendeded that the labels should be defined in pod definition rather than added after pod is created, as labels added later on will be lost in case the pod gets rescheduled. In case of pods defined as part of a deployment, it's best to use the same label as the one defined in the `spec.selector.matchLabels` section of the deployment yaml. The prefix field can be used to avoid naming collision with other labels.
```
edcdavid, line 36:
```suggestion
The goal of this section is to specify the labels to be used to identify the CNF resources under test. It's highly recommended that the labels should be defined in pod definition rather than added after pod is created, as labels added later on will be lost in case the pod gets rescheduled. In case of pods defined as part of a deployment, it's best to use the same label as the one defined in the `spec.selector.matchLabels` section of the deployment yaml. The prefix field can be used to avoid naming collision with other labels.
```

edcdavid: I introduced this trying to find about how to suggest changes

README.md
hamadise: restored the section while removing the non-autodiscovery lines.

edcdavid (approved):

jun-rh (approved):

PR #474, opened 2021-11-12
Use go template for hostPort check

sebrandon1:
```
$ oc get pod busybox1 -o json | jq -r '.spec.containers[0] | select(.ports) | .ports[].hostPort'
null
```

We want to avoid the `null` so switching to go-template:
```
$ oc get pod busybox1 -o go-template='{{range (index .spec.containers 0).ports }}{{if .hostPort}}{{.hostPort}}{{end}}{{end}}'
```

Will review with @greyerof on Monday.

### DIFF
--- a/pkg/tnf/testcases/data/cnf/privilegedpod.go
+++ b/pkg/tnf/testcases/data/cnf/privilegedpod.go
@@ -33,7 +33,7 @@ var PrivilegedPodJSON = string(`{
       "name": "HOST_PORT_CHECK",
       "skiptest": true,
       "loop": 1,
-      "command": "oc get pod %s -n %s -o json | jq -r '.spec.containers[%d] | select(.ports) | .ports[].hostPort'",
+      "command": "oc get pod %s -n %s -o go-template='{{range (index .spec.containers %d).ports }}{{if .hostPort}}{{.hostPort}}{{end}}{{end}}'",
       "action": "allow",
       "expectedstatus": [
         "^(<no value>)*$"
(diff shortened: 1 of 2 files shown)


pkg/tnf/testcases/data/cnf/privilegedpod.go
```diff
@@ -33,7 +33,7 @@ var PrivilegedPodJSON = string(`{
       "name": "HOST_PORT_CHECK",
       "skiptest": true,
       "loop": 1,
-      "command": "oc get pod %s -n %s -o json | jq -r '.spec.containers[%d] | select(.ports) | .ports[].hostPort'",
+      "command": "oc get pod %s -n %s -o go-template='{{range (index .spec.containers %d).ports }}{{if .hostPort}}{{.hostPort}}{{end}}{{end}}'",
```
greyerof, line 36: This will work and fix the issue but in case more than one container have hostPorts, they will be printed without separation like: "80808080". Please add an extra espace char between {{.hostPort}} and the following {{end}} to avoid that. I've tested a more complex template to print the pod's and container's name so it can be debugged with the LOG_LEVEL=trace enabled:
```
go-template='{{$putName := .metadata.name}}{{$cut := (index .spec.containers %d)}}{{range $cut.ports }}{{if .hostPort}}PUT {{$putName}} - CUT {{$cut.name}} has declared hostPort {{.hostPort}}{{\"\\n\"}}{{end}}{{end}}'"
```
The output would be:
```
TRACE  [Nov 15 11:15:48.147][spawner.go: 301] STDOUT for /bin/bash  : PUT test-d75b68dfd-dwkh2 - CUT test has declared hostPort 8080
```

greyerof (approved):

sebrandon1:
Verified with positive and negative cases in minikube.  Related to changes here: https://github.com/test-network-function/cnf-certification-test-partner/pull/59

hamadise (approved):

PR #482, opened 2021-11-18
Optimize test target discovery with multiple namespaces

hamadise:


pkg/config/autodiscover/deployment_info.go
```diff
@@ -113,3 +113,20 @@ func GetTargetDeploymentsByNamespace(namespace string, targetLabel configsection
 
 	return &deploymentList, nil
 }
+
+// GetTargetDeploymentsByLabel will return all deployments that have pods with a given label.
+func GetTargetDeploymentsByLabel(targetLabel configsections.Label) (*DeploymentList, error) {
+	labelQuery := fmt.Sprintf("\"%s\"==\"%s\"", buildLabelName(targetLabel.Prefix, targetLabel.Name), targetLabel.Value)
```
greyerof, line 119: I think this linting problem comes from old code that you copy/pasted to create this new function from the original GetTargetDeploymentsByNamespace(). But that function was updated by the main merge commit you added. Line 102 has the right format string you need here, which is "%q==%q".

hamadise: fixed. Thanks

jun-rh (approved):

greyerof (approved):

PR #483, opened 2021-11-18
Add tests for isMaster/isWorker

sebrandon1:
Add some simple unit tests for coverage of the `isMaster` `isWorker` functions.


greyerof:
I think we missed this from the PR review when this "Node.go" file was introduced, but it should be with all the letters in lower case. Can you update that file as well as your Node_test.go, file, please?

greyerof (approved):

sebrandon1:
@greyerof Okay updated the filenames!  Thanks for the reviews!

jun-rh (approved):

PR #486, opened 2021-11-22
CNFCERT-185:Use imagePullPolicy: IfNotPresent

aabughosh:


pkg/tnf/testcases/data/cnf/gatherfacts.go
```diff
@@ -30,6 +30,17 @@ var GatherPodFactsJSON = string(`{
         "ALLOW_ALL"
       ]
     },
+    {
+      "name": "imagePullPolicy",
+      "skiptest": false,
+      "command": "oc get pod %s -n %s -o json  | jq -r '.spec.containers[0].imagePullPolicy'",
```
greyerof, line 36: With .spec.containers[0] you're only checking the first container of the pod. This should be either in privilegedpod.go (with containers[%d]) or in a new json-based test.

greyerof:
README.md file should also be updated with this new TC/sub-TC.

jun-rh:
General comment:

- This test doesn't seem to be related to access control
- We do not intend to add more tests using this framework. Use the generic handler based on json instead.

PR #487, opened 2021-11-22
Retiring TNF_MINIKUBE_ONLY

jun-rh:


pkg/config/autodiscover/autodiscover_partner.go
```diff
@@ -30,7 +30,7 @@ const (
 )
 
 func IsMinikube() bool {
```
greyerof, line 32: Why not renaming this function to IsNonOcpCluster()?

test-network-function/common/env.go
```diff
@@ -57,7 +57,7 @@ func GetContext() *interactive.Context {
 
 // IsMinikube returns true when the env var is set, OCP only test would be skipped based on this flag
 func IsMinikube() bool {
```
greyerof, line 59: Same.

greyerof (approved):

pkg/config/autodiscover/autodiscover_partner.go
jun-rh: done

sebrandon1 (approved):

hamadise (approved):

PR #488, opened 2021-11-23
CNFCER-185: add handler to check image pull policy

aabughosh:


javierpena:
I noticed the new test is failing because our test pods use "Always" image pull policy instead of "IfNotPresent". I've created PR https://github.com/test-network-function/cnf-certification-test-partner/pull/63 to fix this.

aabughosh:
> I noticed the new test is failing because our test pods use "Always" image pull policy instead of "IfNotPresent". I've created PR [test-network-function/cnf-certification-test-partner#63](https://github.com/test-network-function/cnf-certification-test-partner/pull/63) to fix this.

thanks

test-network-function/lifecycle/suite.go
```diff
@@ -479,3 +482,23 @@ func testOwner(env *config.TestEnvironment) {
...
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestPodDeploymentBestPracticesIdentifier)
+	ginkgo.It(testID, func() {
+		ginkgo.By("Testing image pull policy")
+		context := common.GetContext()
+		for _, podUnderTest := range env.PodsUnderTest {
+			podName := podUnderTest.Name
+			podNamespace := podUnderTest.Namespace
+			podContainerCount := podUnderTest.ContainerCount
+			for i := 0; i < podContainerCount; i++ {
+				tester := imagepullpolicy.NewImagepullpolicy(common.DefaultTimeout, podNamespace, podName, i)
```
jun-rh, line 496: hmm, i thought this one should be an container level test?

jun-rh (changes requested):
what you implemented would pass if each pod has one container with the right image pull policy. i guess we should really check every container.

pkg/tnf/handlers/imagepullpolicy/imagepullpolicy.go
```diff
@@ -0,0 +1,108 @@
...
+	return &reel.Step{
+		Execute: "",
+		Expect:  []string{imagepullpolicy.regex}, // TODO : pass the list of possible regex in here
+		Timeout: imagepullpolicy.timeout,
+	}
+}
+
+// ReelMatch parses the Imagepullpolicy output and set the test result on match.
+func (imagepullpolicy *Imagepullpolicy) ReelMatch(_, _, match string) *reel.Step {
+	// TODO : add the matching logic here and return an appropriate tnf result.
```
jun-rh, line 88: remove todo

pkg/tnf/handlers/imagepullpolicy/imagepullpolicy_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+func createTest() (*tnf.Tester, []reel.Handler, *gojsonschema.Result, error) {
+	values := make(map[string]interface{})
+	values["POD_NAMESPACE"] = testPodNameSpace
+	values["POD_NAME"] = testPodName
+	values["CONTAINER_NUM"] = testContainerNum
+	return generic.NewGenericFromMap(imagepullFilename, pathToTestSchemaFile, values)
+}
+
+func TestLogging_Args(t *testing.T) {
```
jun-rh, line 61: Update function names

pkg/tnf/handlers/imagepullpolicy/imagepullpolicy_test.go
```diff
@@ -17,84 +17,124 @@
...
-func Test_ReelFirstPositive(t *testing.T) {
-	handler := imagepullpolicy.NewImagepullpolicy(testTimeoutDuration, testPodNamespace, testPodName, testContainerCount)
-	assert.NotNil(t, handler)
-	firstStep := handler.ReelFirst()
-	re := regexp.MustCompile(firstStep.Expect[0])
+var (
+	genericTestSchemaFile = path.Join("schemas", "generic-test.schema.json")
+	imagepullFilename     = "imagepullolicy.json"
+	/* #nosec G101 */
+	expectedPassPattern  = "(?m)[1-9]"
```
jun-rh, line 41: these are also copied from logging test?

pkg/tnf/handlers/imagepullpolicy/imagepullpolicy_test.go
```diff
@@ -17,84 +17,124 @@
...
-var (
-	testInputSuccess = "IfNotPresent"
-)
+	assert.Nil(t, err)
+	assert.True(t, jsonParseResult.Valid())
+	assert.NotNil(t, handlers)
+
+	assert.Equal(t, 1, len(handlers))
+	handler := handlers[0]
+	step := handler.ReelMatch(expectedPassPattern, "", testInputSuccess)
```
jun-rh, line 131: does this pass? i think you are using your input with the pattern from the logging test?

test-network-function/lifecycle/suite.go
```diff
@@ -479,3 +485,36 @@ func testOwner(env *config.TestEnvironment) {
 		}
 	})
 }
+
+func testImagePolicy(env *config.TestEnvironment) {
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestPodDeploymentBestPracticesIdentifier)
```
jun-rh, line 490: Need new identifier

pkg/tnf/handlers/imagepullpolicy/imagepullpolicy_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	assert.Nil(t, err)
+	assert.True(t, jsonParseResult.Valid())
+	assert.NotNil(t, handlers)
+
+	assert.Equal(t, 1, len(handlers))
+	handler := handlers[0]
+	step := handler.ReelFirst()
+	expectedCommand := fmt.Sprintf("oc get pod %s -n %s -o json  | jq -r '.spec.containers[%d].imagePullPolicy'", testPodName, testPodNameSpace, testContainerNum)
+	assert.Equal(t, expectedCommand, step.Execute)
```
hamadise, line 91: what would happen if this policy is not defined ?

test-network-function/lifecycle/suite.go
```diff
@@ -479,3 +485,36 @@ func testOwner(env *config.TestEnvironment) {
 		}
 	})
 }
+
+func testImagePolicy(env *config.TestEnvironment) {
```
hamadise, line 489: Three general comments in here : 

1. you're assuming the loop will give you containers belonging to the same pod successively. in golang map, the order is random. so this logic will not work.
2. it's better to get the comment per container directly if possible.
3. the way you're looping here, you'll stop the test in the first failed container. It's better to find all the failed containers. -look at tainted kernel test as reference-

pkg/tnf/handlers/imagepullpolicy/imagepullpolicy_test.go
jun-rh: it seems it's always there when you dump it. it is optional when you create the pod. we found conflicting references about what the default is when it's missing.

jun-rh (dismissed):

test-network-function/lifecycle/suite.go
jun-rh: Nice catch. @aabughosh Just put the original version back, where you loop through podUnderTest.

aabughosh: ohh thanks salah didnt know about that i updated the loop to go into the pods

jun-rh (approved):

hamadise (approved):

PR #490, opened 2021-11-23
add  automount service test

hamadise:


pkg/tnf/handlers/automountservice/automountservice.go
```diff
@@ -0,0 +1,159 @@
...
+
+// WithNamespace specify the namespace
+func WithNamespace(ns string) func(*Automountservice) {
+	return func(as *Automountservice) {
+		as.namespace = ns
+	}
+}
+
+//
+// WithTimeout specify the namespace
```
jun-rh, line 78: copy/paste error

test-network-function/identifiers/identifiers.go
```diff
@@ -162,6 +162,11 @@ var (
 		Url:     formTestURL(common.AccessControlTestKey, "pod-service-account"),
 		Version: versionOne,
 	}
+	//
+	TestPodAutomountServiceAccountIdentifier = claim.Identifier{
+		Url:     formTestURL(common.AccessControlTestKey, "pod-automount-service"),
```
jun-rh, line 167: service => service-account or service-account-token

hamadise: fixed

pkg/tnf/handlers/automountservice/automountservice.go
hamadise: fixed.

jun-rh (approved):

sebrandon1 (approved):

test-network-function/identifiers/identifiers.go
```diff
@@ -617,4 +622,10 @@ the changes for you.`,
 		Remediation:           `make sure containers are not redirecting stdout/stderr`,
 		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 11.1",
 	},
+	TestPodAutomountServiceAccountIdentifier: {
+		Type: normativeResult,
```
jun-rh, line 626: Missing Identifier

pkg/tnf/handlers/automountservice/automountservice.go
```diff
@@ -0,0 +1,160 @@
...
+		Expect:  []string{saRegex},
+		Timeout: as.timeout,
+	}
+}
+
+// ReelMatch parses the Automountservice output and set the test result on match.
+func (as *Automountservice) ReelMatch(_, _, match string) *reel.Step {
+	numExpectedMatches := 2
+	saMatchIdx := 1
+	as.result = tnf.SUCCESS
```
javierpena, line 134: I find the handler logic a bit confusing. If I understood it correctly:

- We will always get to ReelMatch, even if the regex is not matched. So we always set as.result = tnf.SUCCESS (which is a bit confusing). However, that will change with https://github.com/test-network-function/test-network-function/pull/496, so I think that PR could break the behavior in this one.

- Then, we re-check the match string against the pattern, and set as.token accordingly only if the match was found.

- Finally, in accesscontrol/suite.go, we check the .token status of both the service account and pod, and that is where we actually decide if the test passes or fails.

Is this correct? If so, it would be great if you could add some comment in the handler code to explain that behavior, otherwise it could be hard to debug in the future. Also, you may need to adjust the pattern to match against if PR 496 is merged, since it will always fail for pods without an explicit automountServiceAccountToken key set.

test-network-function/accesscontrol/suite.go
```diff
@@ -272,24 +272,66 @@ func testRoles(env *config.TestEnvironment) {
...
 			test.RunAndValidate()
-			serviceAccountName := tester.GetServiceAccountName()
-			gomega.Expect(serviceAccountName).ToNot(gomega.BeEmpty())
+			serviceAccountToken := tester.Token()
+			tester = automountservice.NewAutomountservice(automountservice.WithNamespace(podNamespace), automountservice.WithPodname(podName))
+			test, err = tnf.NewTest(context.GetExpecter(), tester, []reel.Handler{tester}, context.GetErrorChannel())
+			gomega.Expect(err).To(gomega.BeNil())
+			test.RunAndValidate()
+			podToken := tester.Token()
+			// the test would pass iif token is explicitly set to false
```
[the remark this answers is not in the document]
hamadise: iif == if and only if

pkg/tnf/handlers/automountservice/automountservice.go
```diff
@@ -0,0 +1,160 @@
...
+	as := &Automountservice{
+		timeout: common.DefaultTimeout,
+		result:  tnf.ERROR,
+		token:   TokenNotSet,
+	}
+	for _, o := range options {
+		o(as)
+	}
+	if as.podCheck {
+		as.args = []string{"oc", "-n", as.namespace, "get", "pods", as.podname, "-o", "json", "|", "jq", "-r", ".spec"}
```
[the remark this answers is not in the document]
hamadise: fixed.

pkg/tnf/handlers/automountservice/automountservice.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+const (
+	saRegex = `(?m)"automountServiceAccountToken": (.+)`
+	False   = "false,"
+	True    = "true,"
+)
+
+// NewAutomountservice returns a new Automountservice handler struct.
+func NewAutomountservice(options ...func(*Automountservice)) *Automountservice {
```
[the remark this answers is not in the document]
hamadise: fixed with capitalization

pkg/tnf/handlers/automountservice/automountservice.go
hamadise: your understanding is correct. I agree that it's bit confusing, but what you described is the logic of most of our tests. 

1- define the command. 
2- run in using generic handlers (via JSON) or using a new handler
3- when running using generic handlers, the status is all that matters and we get the results right away.
4- when choosing new handler, we can return the status or some other data that the test in the suite decides if there's a failure or not. 

*) this specific case here is not intuitive since the result of the Test depends on the results of two separate calls (or handlers)

The part that should be clear is where we decide if there's a failure or not in the suite.go

hamadise: added more comments to make the test clearer in suite.go

test-network-function/identifiers/identifiers.go
hamadise: fixed

pkg/tnf/handlers/automountservice/automountservice.go
javierpena: Thanks for the explanation. It looks much clearer now, with the explanations and the initial match against allRegex in the step.

test-network-function/accesscontrol/suite.go
```diff
@@ -264,24 +264,80 @@ func testRoles(env *config.TestEnvironment) {
...
+}
+
+//nolint:funlen
+func testAutomountService(env *config.TestEnvironment) {
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestPodAutomountServiceAccountIdentifier)
+	ginkgo.It(testID, func() {
+		ginkgo.By("Should have automountServiceAccountToken set to false")
+		msg := []string{}
+		for _, podUnderTest := range env.PodsUnderTest {
+			ginkgo.By(fmt.Sprintf("check the existence of pod service account %s %s", podUnderTest.Namespace, podUnderTest.Name))
```
greyerof, line 289: Minor fix for readability: "... service account %s %s" -> "service account for pod %s (ns %s), podUnderTest.Name, podUnderTest.Namespace"

test-network-function/accesscontrol/suite.go
```diff
@@ -264,24 +264,80 @@ func testRoles(env *config.TestEnvironment) {
...
+				msg = append(msg, fmt.Sprintf("serviceaccount %s:%s is configured with automountServiceAccountToken set to true, impacting pod %s ", podNamespace, serviceAccountName, podName))
+			}
+			// the token should be set explicitly to false, otherwise, it's a failure
+			// register the error message and check the next pod
+			if serviceAccountToken == automountservice.TokenNotSet {
+				msg = append(msg, fmt.Sprintf("serviceaccount %s:%s is not configured with automountServiceAccountToken set to false, impacting pod %s ", podNamespace, serviceAccountName, podName))
+			}
+		}
+		if len(msg) > 0 {
+			_, err := ginkgo.GinkgoWriter.Write([]byte(strings.Join(msg, "")))
```
greyerof, line 335: Suggestion: use "tnf.ClaimFilePrintf()"

greyerof (approved):
Minor comments that can be addressed in different PR if needed.

test-network-function/accesscontrol/suite.go
hamadise: fixed

PR #493, opened 2021-11-25
Negative test

shimritproj:
Proceed to task 166, creating a negative test
"Access Control-Host-Resource-CAPABILITY_CHECK"


pkg/tnf/handlers/container/pod.go
```diff
@@ -100,7 +100,8 @@ func (p *Pod) ReelMatch(_, _, match string) *reel.Step {
 			p.result = tnf.SUCCESS
 			return nil
 		}
-		replacer := strings.NewReplacer(`[`, ``, "\"", ``, `]`, ``, `, `, `,`)
+		replacer := strings.NewReplacer(`[`, ``, "\"", ``, `]`, ``, `, `, `,`, "\n", ``, "\t", ``, ` `, ``)
```
jun-rh, line 103: ``, `, ` `` is no longer needed

greyerof (approved):
The last replacer pair will also remove space chars in the middle of the output. Not sure if this is what we want, but we can leave this as technical-debt.

pkg/tnf/handlers/container/pod.go
jun-rh: Actually, Gonzalo brought up a good point. Removing space this way is a bit too much. @shimritproj Let's remove both
````, `, `,` and ` `, ``
do this instead:
```suggestion
		if contains(p.ExpectStatus, strings.Trim(status, " ")) {
```

shimritproj: @jun-rh  @greyerof right, I changed.

sebrandon1 (approved):

greyerof (approved):

jun-rh (approved):

PR #496, opened 2021-11-29
Fix for ReelMatch() always getting called

jun-rh:


pkg/tnf/reel/reel.go
```diff
@@ -206,32 +205,31 @@ func (r *Reel) Step(step *Step, handler Handler) error {
...
 				}
 				match, matchStatus := r.stripEmulatedPromptFromOutput(result.Match[0])
 				log.Debugf("command status: output=%s, match=%s, outputStatus=%d, matchStatus=%d", output, match, outputStatus, matchStatus)
 
-				matchIndex := strings.Index(output, match)
-				var before string
-				// special case:  the match regex may be nothing at all.
-				if matchIndex > 0 {
-					before = output[0 : matchIndex-1]
+				if result.CaseIdx != len(batcher[0].Cases())-1 {
```
greyerof, line 215: This is a good time to rename the batcher var to "batchers", since it's a slice. Also, this deserves a comment to remind the reader that we added the "empty" batcher as the last Case in the batcher, so that's why we need to filter it out.

greyerof, line 215: batcher[0] has the Args() BSnd batcher, which is used to send the handler's Args() command. You should use batcher[len(batcher) - 1] or similar, to avoid handler's that don't have an Args() command.

sebrandon1 (approved):

edcdavid:
Not due to this PR, but noticed that the current Idx might be wrong in the unit tests (test_test.go). Idx should be 1 to point to the batch expect batcher since the unit test is using a exec command (ls)

			{
				Idx:    0,
				Output: fakeSentinelOutput(),
				Match:  []string{fakeSentinelOutput()}},
		},

edcdavid (approved):

greyerof (approved):

jun-rh:
> Not due to this PR, but noticed that the current Idx might be wrong in the unit tests (test_test.go). Idx should be 1 to point to the batch expect batcher since the unit test is using a exec command (ls)
> 
> ```
> 		{
> 			Idx:    0,
> 			Output: fakeSentinelOutput(),
> 			Match:  []string{fakeSentinelOutput()}},
> 	},
> ```

Thanks for the reminder. The unit test was due for an update for not only this change, but also changes on return code processing. Just pushed a commit for that.

PR #499, opened 2021-12-01
Fixing Multus skip logic

edcdavid:
Skipping all container with no multus IPs
If no container available in the end, skipping the test


javierpena:
LGTM, and the logic makes sense.

PR #500, opened 2021-12-01
Insert claim.json file in results HTML page

javierpena:
To simplify log viewing, we want to have the CI job insert the
claim.json file automatically inside the results.html page. This cannot
be done by dynamically loading it due to CORS restrictions, but we can
convert the JSON file into a javascript file, then load it from the HTML
page.

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -153,7 +153,8 @@ jobs:
           path: |
             test-network-function/*.xml
             test-network-function/claim.json
-            script/results.html
+            test-network-function/claimjson.js
+            test-network-function/results.html
 
       # Perform smoke tests using a TNF container.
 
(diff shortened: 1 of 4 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -190,6 +199,7 @@ jobs:
           path: |
             ${{ env.TNF_OUTPUT_DIR }}/*.xml
             ${{ env.TNF_OUTPUT_DIR }}/claim.json
+            script/claimjson.js
```
jun-rh, line 202: Does it make more sense to copy the html and js file to the output dir? Also, why not do this in run-cnf-suites.sh so that it happens for any type of test run (github, dev, container)?

javierpena:
Currently, the script does what it should when run locally, but not when run via run-tnf-container.sh. I'm checking why.

.github/workflows/pre-main.yaml
jun-rh: It's just because you made it too pretty :)

javierpena:
The latest commit fixes output in the container test, I think it's ready to go.

jun-rh (approved):

edcdavid (approved):
Nice work !

PR #502, opened 2021-12-02
Ensure drained nodes are uncordoned if test times out

javierpena:
In the pod recreation test, when deployments are not restarted in time
after a node is drained, that node would remain drained forever and
never be uncordoned.

We can defer the node uncordon operation, so it will always be executed.
If the node was successfully uncordoned, a second uncordon will become a
no-op.

We also remove some unneeded checks after the call to uncordonNode(),
since they are already performed by waitForAllDeploymentsReady(), and
move the final waitForAllDeploymentsReady() check outside of the loop,
so we only do it once.

### DIFF
--- a/test-network-function/lifecycle/suite.go
+++ b/test-network-function/lifecycle/suite.go
@@ -316,12 +316,7 @@ func testNodeDrain(env *config.TestEnvironment, nodeName string) {
 		waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
 	}
 	// If we got this far, all deployments are ready after draining the node
-	msg := fmt.Sprintf("Node drain for %s succeeded\n", nodeName)
-	log.Info(msg)
-	_, err := ginkgo.GinkgoWriter.Write([]byte(msg))
-	if err != nil {
-		log.Errorf("Ginkgo writer could not write because: %s", err)
-	}
+	tnf.ClaimFilePrintf("Node drain for %s succeeded\n", nodeName)
 }
 
 func testPodsRecreation(env *config.TestEnvironment) {
--- a/test-network-function/lifecycle/suite.go
+++ b/test-network-function/lifecycle/suite.go
@@ -401,6 +401,7 @@ func drainNode(node string) {
 	gomega.Expect(err).To(gomega.BeNil())
 	result, err := test.Run()
 	if err != nil || result == tnf.ERROR {
+		uncordonNode(node) // Try to undo the node drain
 		log.Fatalf("Test skipped because of draining node failure - platform issue")
 	}
 }
--- a/test-network-function/lifecycle/suite.go
+++ b/test-network-function/lifecycle/suite.go
@@ -300,6 +300,19 @@ func shutdownTest(podNamespace, podName string) {
 	test.RunAndValidate()
 }
 
+func testNodeDrain(env *config.TestEnvironment, nodeName string) {
+	// We need to delete all Oc sessions because the drain operation is often deleting oauth-openshift pod
+	// This results in lost connectivity for oc sessions
+	env.ResetOc()
+	// drain node
+	drainNode(nodeName)
+	// Ensure the node is uncordoned before exiting the function
+	defer uncordonNode(nodeName)
+	for _, ns := range env.NameSpacesUnderTest {
+		waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
+	}
+}
+
 func testPodsRecreation(env *config.TestEnvironment) {
 	deployments := make(dp.DeploymentMap)
 	var notReadyDeployments []string
@@ -329,20 +342,10 @@ func testPodsRecreation(env *config.TestEnvironment) {
 				log.Debug("node ", n.Name, " has no deployment, skip draining")
 				continue
 			}
-			// We need to delete all Oc sessions because the drain operation is often deleting oauth-openshift pod
-			// This result in lost connectivity oc sessions
-			env.ResetOc()
-			// drain node
-			drainNode(n.Name) // should go in this
-			// Ensure the node is uncordoned even if we have a failure in waitForAllDeploymentsReady
-			defer uncordonNode(n.Name)
-			for _, ns := range env.NameSpacesUnderTest {
-				waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
-			}
-			uncordonNode(n.Name)
+			testNodeDrain(env, n.Name)
 		}
+		// Once all tests are done, wait for all deployments to be ready. Otherwise, pods might be unreacheable during the next discovery
 		for _, ns := range env.NameSpacesUnderTest {
-			// wait for all deployment to be ready otherwise, pods might be unreacheable during the next discovery
 			waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
 		}
 	})
--- a/test-network-function/lifecycle/suite.go
+++ b/test-network-function/lifecycle/suite.go
@@ -316,7 +316,7 @@ func testNodeDrain(env *config.TestEnvironment, nodeName string) {
 		waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
 	}
 	// If we got this far, all deployments are ready after draining the node
-	tnf.ClaimFilePrintf("Node drain for %s succeeded\n", nodeName)
+	tnf.ClaimFilePrintf("Node drain for %s succeeded", nodeName)
 }
 
 func testPodsRecreation(env *config.TestEnvironment) {
--- a/test-network-function/lifecycle/suite.go
+++ b/test-network-function/lifecycle/suite.go
@@ -124,7 +124,7 @@ func waitForAllDeploymentsReady(namespace string, timeout, pollingPeriod time.Du
 	var notReadyDeployments []string
 
 	for elapsed < timeout {
-		_, notReadyDeployments := getDeployments(namespace)
+		_, notReadyDeployments = getDeployments(namespace)
 		log.Debugf("Waiting for deployments to get ready, remaining: %d deployments", len(notReadyDeployments))
 		if len(notReadyDeployments) == 0 {
 			break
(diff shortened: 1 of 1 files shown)


test-network-function/lifecycle/suite.go
```diff
@@ -334,21 +334,16 @@ func testPodsRecreation(env *config.TestEnvironment) {
...
-				// verify deployments are ready again
-				_, notReadyDeployments = getDeployments(ns)
-				if len(notReadyDeployments) > 0 {
-					uncordonNode(n.Name)
-					ginkgo.Fail(fmt.Sprintf("did not create replicas when node %s is drained", n.Name))
-				}
 			}
 			uncordonNode(n.Name)
-
-			for _, ns := range env.NameSpacesUnderTest {
```
jun-rh, line 348: Shouldn't this be part of the defer block?

test-network-function/lifecycle/suite.go
```diff
@@ -334,21 +334,16 @@ func testPodsRecreation(env *config.TestEnvironment) {
...
-				// wait for all deployment to be ready otherwise, pods might be unreacheable during the next discovery
-				waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
-			}
+		}
+		for _, ns := range env.NameSpacesUnderTest {
+			// wait for all deployment to be ready otherwise, pods might be unreacheable during the next discovery
+			waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
```
jun-rh, line 346: Shouldn't this be part of the defer block?

test-network-function/lifecycle/suite.go
```diff
@@ -334,21 +334,16 @@ func testPodsRecreation(env *config.TestEnvironment) {
 			env.ResetOc()
 			// drain node
 			drainNode(n.Name) // should go in this
+			// Ensure the node is uncordoned even if we have a failure in waitForAllDeploymentsReady
+			defer uncordonNode(n.Name)
```
jun-rh, line 338: It gets a bit tricky to do defer inside a loop. Easier if the loop body becomes a function with a defer block.

test-network-function/lifecycle/suite.go
javierpena: Not really, based on the comment it's a post-recovery action to make sure the environment is clean before moving on.

I guess we could remove it, since we're doing it already after draining each node, but in some cases it might be good to keep it in case some operator tries to re-distribute the load after enabling a node.

test-network-function/lifecycle/suite.go
javierpena: Ok, I think I got the idea. I'll add it in the next version.

greyerof (approved):

test-network-function/lifecycle/suite.go
```diff
@@ -329,26 +342,11 @@ func testPodsRecreation(env *config.TestEnvironment) {
 				log.Debug("node ", n.Name, " has no deployment, skip draining")
 				continue
 			}
-			// We need to delete all Oc sessions because the drain operation is often deleting oauth-openshift pod
-			// This result in lost connectivity oc sessions
-			env.ResetOc()
```
jun-rh, line 334: @edcdavid You added this. Not sure why it's inside the loop?

edcdavid: This is to prevent the broken sessions issues. we close all the sessions before they are potentially  disconnected

edcdavid: Missed your initial point. Yes, this call should be made only once before starting to drain all the nodes.

test-network-function/lifecycle/suite.go
```diff
@@ -300,6 +300,19 @@ func shutdownTest(podNamespace, podName string) {
...
 }
 
+func testNodeDrain(env *config.TestEnvironment, nodeName string) {
+	// We need to delete all Oc sessions because the drain operation is often deleting oauth-openshift pod
+	// This results in lost connectivity for oc sessions
+	env.ResetOc()
+	// drain node
+	drainNode(nodeName)
+	// Ensure the node is uncordoned before exiting the function
+	defer uncordonNode(nodeName)
```
jun-rh, line 310: i think we need to call the wait function again as part of the defer block

test-network-function/lifecycle/suite.go
```diff
@@ -300,6 +300,19 @@ func shutdownTest(podNamespace, podName string) {
 	test.RunAndValidate()
 }
 
+func testNodeDrain(env *config.TestEnvironment, nodeName string) {
+	// We need to delete all Oc sessions because the drain operation is often deleting oauth-openshift pod
+	// This results in lost connectivity for oc sessions
+	env.ResetOc()
+	// drain node
+	drainNode(nodeName)
+	// Ensure the node is uncordoned before exiting the function
```
jun-rh, line 309: still need the checking here so we know if pods get recreated properly, i.e. test pass or fail

javierpena: I'm not sure about that. waitForAllDeploymentsReady() is already checking if all pods are already recreated, and failing if they're not. So that status check was redundant, and we were never reaching it in case of a timeout.

test-network-function/lifecycle/suite.go
javierpena: Why should we? If waitFor..() succeds, we don't need to run it again, and if it fails, it will fail twice if we call it inside the defer block.

We have an extra waitFor...() call outside the for loop, in case there is some operator moving pods around when we uncordon the last node, shouldn't that be enough?

test-network-function/lifecycle/suite.go
javierpena: Should I move it outside the loop, then?

edcdavid: yes

test-network-function/lifecycle/suite.go
jun-rh: no, the one outside should be removed as it's not reliable, it may be skipped if we panic. we do need to do it at least once in the loop so we know if the test result is success or failure and another time in the fail/error cases after uncordon. Putting it in defer would make it do the second one in all cases, but it shouldn't hurt in success case.

edcdavid, line 310: Add a "uncordonNode(node)" on line 383 to uncordon the node if the drain command fails

edcdavid (approved):

test-network-function/lifecycle/suite.go
javierpena: Aha, I got it now. Thanks for the explanation! I'll update the PR to reflect this.

edcdavid: Yes, it is in the drainNode function. If the drain command fails, the node will remain cordoned
func drainNode(node string) {
	context := common.GetContext()
	tester := dd.NewDeploymentsDrain(drainTimeout, node)
	result, err := test.Run()
	if err != nil || result == tnf.ERROR {
                *uncordonNode(node)*
		log.Fatalf("Test skipped because of draining node failure - platform issue")
	}
}

sebrandon1 (approved):

test-network-function/lifecycle/suite.go
```diff
@@ -300,6 +300,22 @@ func shutdownTest(podNamespace, podName string) {
...
+	// Ensure the node is uncordoned before exiting the function,
+	// and all deployments are ready
+	defer func() {
+		uncordonNode(nodeName)
+		for _, ns := range env.NameSpacesUnderTest {
+			waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
+		}
+	}()
+	for _, ns := range env.NameSpacesUnderTest {
+		waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
```
jun-rh, line 315: need to check and report test result after this wait

test-network-function/lifecycle/suite.go
```diff
@@ -300,6 +300,22 @@ func shutdownTest(podNamespace, podName string) {
...
 
+func testNodeDrain(env *config.TestEnvironment, nodeName string) {
+	// drain node
+	drainNode(nodeName)
+	// Ensure the node is uncordoned before exiting the function,
+	// and all deployments are ready
+	defer func() {
+		uncordonNode(nodeName)
+		for _, ns := range env.NameSpacesUnderTest {
+			waitForAllDeploymentsReady(ns, scalingTimeout, scalingPollingPeriod)
```
jun-rh, line 311: Need to check if the deployments actually become ready again. If not, I think we can abort the test run. Otherwise it would just create more problematic test execution logs for us to triage with no real benefit

test-network-function/lifecycle/suite.go
javierpena: Done already.

test-network-function/lifecycle/suite.go
javierpena: How can we do that? waitForAllDeploymentsReady() will fail if some deployment is not ready, so the rest of the defer function will not be executed.

I see there's ginkgo.AbortSuite() in 2.0.0 [1], but we're still in pre-2.0. Even calling panic() would just fail this test, but nothing else.

[1] - https://github.com/onsi/ginkgo/issues/734

jun-rh: How about extracting this inline function out and have another defer in that funciton?

jun-rh (approved):

greyerof (approved):

test-network-function/lifecycle/suite.go
```diff
@@ -135,7 +143,11 @@ func restoreDeployments(env *config.TestEnvironment) {
 
 		if len(notReadyDeployments) > 0 {
 			// Wait until the deployment is ready
-			waitForAllDeploymentsReady(deployment.Namespace, scalingTimeout, scalingPollingPeriod)
+			notReady := waitForAllDeploymentsReady(deployment.Namespace, scalingTimeout, scalingPollingPeriod)
+			if notReady != 0 {
+				collectNodeAndPendingPodInfo(deployment.Namespace)
```
jun-rh, line 148: @greyerof I think we missed this? Aborting here would make the next if block useless. With your original design, we should only do this after the last attempt of restoring the replicas?

greyerof: If I'm not wrong, in both the original and current restoreDeployments, the next "if block" won't be reached if waitForAllDeploymentsReady on line 146 fails. The difference is that in the original version, it was implemented with  gomega.Eventually, which makes the TC to fail but the TS continues, while this log.Fatalf() will stop the whole TS/tnf.

PR #504, opened 2021-12-02
Add more unit tests for autodiscover

sebrandon1:
Added a test for `FindTestCrdNames` function.  I've got an idea about how to tackle the other functions in the `autodiscover` package but I just wanted to get this one locked in first.

The GetContext() function expects a `SHELL` environment variable set to determine how to make a new session.  I have added it to the GitHub workflows.


Makefile
```diff
@@ -14,6 +14,8 @@
 # Tasks provide shortcuts to common operations that occur frequently during
 # development. This includes running configured linters and executing unit tests
 
+SHELL = /bin/bash
```
jun-rh, line 17: would this cause problems if someone is not using bash? maybe set it in the workflow only?

sebrandon1: Done.  Switched it to the github workflow.

jun-rh (approved):

.github/workflows/merge.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
 
       - name: Run Tests
         run: make test
+        env:
+          SHELL: /bin/bash
```
greyerof, line 29: Why is this env var needed for this unit test?

sebrandon1: When the tests are running and attempt to:
```
// GetContext spawns a new shell session and returns its context
func GetContext(verbose bool) *Context {
	context, err := SpawnShell(CreateGoExpectSpawner(), defaultTimeout, Verbose(verbose), SendTimeout(defaultTimeout))
	if err != nil || context == nil || context.GetExpecter() == nil {
		log.Panicf("can't get a proper context for test execution")
	}
	return context
}
```

It errors out saying it can't get a proper context because it doesn't know what SHELL environment to use.  We have to be explicit in the env running the test.

greyerof: Thanks!

greyerof (approved):
Please add why the envar is needed in the PR description so it appears in the final merge commit.

sebrandon1:
Done!

PR #506, opened 2021-12-05
CNFCERT-178: ping need to be skiiped if there is no ping

aabughosh:


pkg/utils/utils.go
```diff
@@ -77,7 +84,7 @@ func escapeToJSONstringFormat(line string) (string, error) {
 
 // ExecuteCommand uses the generic command handler to execute an arbitrary interactive command, returning
 // its output wihout any other check.
-func ExecuteCommand(command string, timeout time.Duration, context *interactive.Context, failureCallbackFun func()) string {
+func ExecuteCommand(command string, timeout time.Duration, context *interactive.Context, failureCallbackFun func(), runopt string) string {
```
hamadise, line 87: I don't think this function should have extra parameters, but if it's necessary. It's better to create an Enum and pass that Enum. Right now, you're passing a string, and not using it, you only care if it's empty or not.

pkg/utils/utils.go
```diff
@@ -26,6 +27,12 @@ var (
 	handlerJSONSchemaFilePath = path.Join(pathRelativeToRoot, "schemas", "generic-test.schema.json")
 )
 
+const (
+	sucssesmsg = "sucsses to run the test\n"
```
javierpena, line 31: There's a typo here, I guess you meant "successmsg" and "success"

pkg/utils/utils.go
```diff
@@ -93,9 +100,28 @@ func ExecuteCommand(command string, timeout time.Duration, context *interactive.
...
+		var message string
+		test.RunWithCallbacks(func() {
+			testok = true
+			message = sucssesmsg
+		}, func() {
+			message = failmsg
+		}, func(e error) {
+			message = errmsg
+		})
+		_, err = ginkgo.GinkgoWriter.Write([]byte(message))
```
javierpena, line 117: You may want to use tnf.ClaimFilePrintf(), which is a helper that does GinkgoWriter.Write() with associated error handling for you.

test-network-function/networking/suite.go
```diff
@@ -80,26 +80,31 @@ func testDefaultNetworkConnectivity(env *config.TestEnvironment, count int) {
...
+					testPing(testOrchestrator.Oc, cut.DefaultNetworkIPAddress, count)
+					ginkgo.By(fmt.Sprintf("a Ping is issued from %s(%s) to %s(%s) %s", cut.Oc.GetPodName(),
+						cut.Oc.GetPodContainerName(), testOrchestrator.Oc.GetPodName(), testOrchestrator.Oc.GetPodContainerName(),
+						testOrchestrator.DefaultNetworkIPAddress))
+					testPing(context, testOrchestrator.DefaultNetworkIPAddress, count)
+				}
 			}
-			if !found {
-				ginkgo.Skip("No container found suitable for connectivity test")
+			if !found || !containerping {
```
javierpena, line 106: I'd prefer to have two separate checks here. One for "if !found" and another one for "if !containerping". That would allow us to be specific with the Skip message when debugging.

jun-rh:
David's nsenter change is in. This is no longer needed.

PR #512, opened 2021-12-07
Utility function to collect status on lifecycle test failure

jun-rh:


test-network-function/lifecycle/suite.go
```diff
@@ -375,6 +375,17 @@ func getDeployments(namespace string) (deployments dp.DeploymentMap, notReadyDep
...
 }
 
+//nolint:deadcode // to be used in Javier's change
+func collectNodeAndPendingPodInfo(ns string) {
+	context := common.GetContext()
+	nodeStatus, _ := utils.ExecuteCommand("oc get nodes -o json | jq '.items[]|{name:.metadata.name, taints:.spec.taints}'", common.DefaultTimeout, context)
+	common.TcClaimLogPrintf("Namespace: %s\nNode status:\n%s", ns, nodeStatus)
+	podStatus, _ := utils.ExecuteCommand("oc get pods -n "+ns+" --field-selector status.phase!=Running -o json | jq '.items[]|{name:.metadata.name, status:.status}'", common.DefaultTimeout, context)
+	common.TcClaimLogPrintf("Pending Pods:\n%s", podStatus)
+	events, _ := utils.ExecuteCommand("oc get events -n "+ns+" -o json --sort-by='.lastTimestamp' | jq '.items[]|{object:.involvedObject, reason:.reason, type:.type, message:.message, lastSeen:.lastTimestamp}'", common.DefaultTimeout, context)
```
sebrandon1, line 385: Would it be better to build these strings with `fmt.Sprintf` instead of `+ns+` ?

test-network-function/lifecycle/suite.go
```diff
@@ -375,6 +375,17 @@ func getDeployments(namespace string) (deployments dp.DeploymentMap, notReadyDep
 	return deployments, notReadyDeployments
 }
 
+//nolint:deadcode // to be used in Javier's change
+func collectNodeAndPendingPodInfo(ns string) {
+	context := common.GetContext()
+	nodeStatus, _ := utils.ExecuteCommand("oc get nodes -o json | jq '.items[]|{name:.metadata.name, taints:.spec.taints}'", common.DefaultTimeout, context)
+	common.TcClaimLogPrintf("Namespace: %s\nNode status:\n%s", ns, nodeStatus)
+	podStatus, _ := utils.ExecuteCommand("oc get pods -n "+ns+" --field-selector status.phase!=Running -o json | jq '.items[]|{name:.metadata.name, status:.status}'", common.DefaultTimeout, context)
```
edcdavid, line 383: Also exclude phase=Succeeded from the results

test-network-function/lifecycle/suite.go
edcdavid, line 385: Should we select only failures here as well? Add:
--field-selector type!=Normal

jun-rh: Good point

test-network-function/lifecycle/suite.go
jun-rh: Right

edcdavid (approved):

hamadise (approved):

PR #515, opened 2021-12-09
Observability: Continue TCs with other CUTs/CRDs when someone fails.

greyerof:


test-network-function/observability/suite.go
```diff
@@ -69,32 +70,46 @@ var _ = ginkgo.Describe(common.ObservabilityTestKey, func() {
 func testLogging() {
 	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestLoggingIdentifier)
 	ginkgo.It(testID, func() {
+		failedCutIds := []*configsections.ContainerIdentifier{}
 		for _, cut := range env.ContainersUnderTest {
-			ginkgo.By(fmt.Sprintf("Test container: %+v. should emit at least one line of log to stderr/stdout", cut.ContainerIdentifier))
-			loggingTest(cut.ContainerIdentifier)
+			cutIdentifier := &cut.ContainerIdentifier
+			ginkgo.By(fmt.Sprintf("Test container: %+v. should emit at least one line of log to stderr/stdout", cutIdentifier))
```
javierpena, line 76: Maybe we could keep the test in a separate function (loggingTest), and just make it return the slice of failedCutIds? That would make it a bit more readable.

greyerof: So we have different views on this :)
I removed the old loggingTest because I thought it was weird having an almost empty function "testLogging" function calling another one called "loggingTest" with so few lines.  Also, the content of the TC seemed to me relatively easy to follow, since it's just instantiating the handler tester per container and adding the failed ones to the array.

jun-rh (approved):

test-network-function/observability/suite.go
javierpena: Ok, no worries then.

javierpena (approved):

PR #511, opened 2021-12-07
Network connectivity test based on running nsenter/ping on debug pods

edcdavid:
Removing dependency on partner pod to run tests (networking)
- using nsenter command to execute any host (openshift/minikube node ) command into a container namespace. Added commands to collect container UID and PID and run the command (nsenter or normal) on nodes
- modified the ping handler to use nsenter to ping from the container namespace. CNF no longer require to have ping. We can use this example for other similar tests relying on commands in the container.
- modified the networking tests to not rely on partner pod:
  -  instead of the partner pod pinging the CNF interfaces, the tnf suite picks a container per network to initiate the ping. This is done for the default pod network, and each multus network attachment. The selected container must have an interface in the network it is supposed to ping to be selected
  - for each network, a list a target containers having interfaces in the same subnet is compiled. these are the containers that will be tested by ping 
  - ping is bi-directional, so no need to originate ping from the target container

This PR needs daemonset support in partner repo to pass tests:
  - https://github.com/test-network-function/cnf-certification-test-partner/pull/66
  - https://github.com/test-network-function/cnf-certification-test-partner/pull/67

Will update documentation and remove partner pod in a separate PR

PR #516, opened 2021-12-09
Refactoring oc session access for recreating on demand

jun-rh:


hamadise (approved):

pkg/config/config.go
```diff
@@ -59,18 +59,31 @@ func getConfigurationFilePathFromEnvironment() string {
 // as the reference to the interactive.Oc instance, the reference to the test configuration, and the default network
 // IP address.
 type Container struct {
-	ContainerConfiguration  configsections.ContainerConfig
-	Oc                      *interactive.Oc
+	configsections.ContainerConfig
```
hamadise, line 62: do you need a variable here ?

jun-rh: i did it on purpose so test code is more readable, e.g. cut.PodName instead of cut.ContainerConfig.ContainerIdentifier.PodName

jun-rh: Also take a look at the latest commit. I redefined HasDebugPod() so I don't have to do nil check everywhere just to avoid panic on minikube

pkg/config/config.go
```diff
@@ -59,18 +59,31 @@ func getConfigurationFilePathFromEnvironment() string {
 // as the reference to the interactive.Oc instance, the reference to the test configuration, and the default network
 // IP address.
 type Container struct {
-	ContainerConfiguration  configsections.ContainerConfig
```
edcdavid, line 62: ContainerConfiguration is used by:
https://github.com/test-network-function/test-network-function/pull/511
Why not keep it ?

edcdavid: ok, I see this is an anonymous field, cool. 
http://golangtutorials.blogspot.com/2011/06/anonymous-fields-in-structs-like-object.html

edcdavid (approved):

PR #517, opened 2021-12-10
Close/Kill shell sessions on oc.Close()

greyerof:
- Kills the underlying shell session when oc.Close() is called.
~~- Wait 1 sec at the end of env.ResetOc() functions to allow channels to
  propagate the "done" signals. This is needed to avoid race conditions
with the "oc adm drain" command, which can be executed first, evicting
the pods whose shell sessions' expecters haven't processed the "done"
signal yet.~~
- Minor refactor in the stdout/stderr mirror pipes to detect EOF as
  normal way to return.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -245,6 +245,9 @@ func (env *TestEnvironment) ResetOc() {
 		cut.Oc.Close()
 		cut.Oc = nil
 	}
+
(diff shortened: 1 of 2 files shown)


greyerof:
Closing this PR because the close/kill method was implemented by this PR: https://github.com/test-network-function/test-network-function/pull/565

PR #518, opened 2021-12-10
removing dependency on ip utility (via nsenter)

edcdavid:
Using the nsenter utility to run the "ip" command from the debug pod in the container namespace. containers that do not have the executable can pass networking tests.
The createContainers() function was split into:
- createContainersNoIP(): populates the container list but does not get the default IP from the containers. Before we were getting the ip from the container oc, now we need to run nsenter with the node oc. But at this moment the list of nodes is not ready yet since it relies on the nodes selected by the CNF replicaset
- recordContainersDefaultIP(): This second part is getting the container default IP after the node and container lists are established using nsenter and node oc.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -29,6 +29,7 @@ import (
 	"github.com/test-network-function/test-network-function/pkg/tnf/handlers/ipaddr"
 	"github.com/test-network-function/test-network-function/pkg/tnf/interactive"
 	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
+	"github.com/test-network-function/test-network-function/pkg/utils"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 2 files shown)


jun-rh:
@hamadise I don't see a way around this two-step approach unless we redesign the node part, which wouldn't be part of 3.2. We can merge this if you feel the same way.

hamadise:
> @hamadise I don't see a way around this two-step approach unless we redesign the node part, which wouldn't be part of 3.2. We can merge this if you feel the same way.

Right. The easiest way for now is to merge this. We don't know the nodes and we don't want to open sessions with all of them before, so the IP discovery is not possible for now. 
I think we can merge it.

hamadise (approved):

jun-rh (approved):

PR #519, opened 2021-12-13
Enable diagnostic suite tests in non-OCP environments

javierpena:
Since [1] was merged, we can run the diagnostic suite tests in non-OCP
environments, like minikube. Let's enable them.

[1] - https://github.com/test-network-function/cnf-certification-test-partner/pull/67


javierpena:
It looks like the container-based versions of minikube lack some utils that are present in the VM-based versions (jq, ifconfig). I'm looking into it.

javierpena:
Keeping the WIP label until https://github.com/test-network-function/cnf-certification-test-partner/pull/72 is merged.

javierpena:
> It shouldn't be all empty?
> 
> ```
>         "Master": {
>           "NodeName": "",
>           "Lscpu": null,
>           "Ifconfig": null,
>           "Lsblk": null,
>           "Lspci": null
>         },
>         "Worker": {
>           "NodeName": "",
>           "Lscpu": null,
>           "Ifconfig": null,
>           "Lsblk": null,
>           "Lspci": null
>         }
>       }```
> ```

This is happening because we first run the diagnostic suite alone, then the rest: https://github.com/test-network-function/test-network-function/blob/main/.github/workflows/pre-main.yaml#L144-L148

I don't see a reason to do that, should we just run all in one go?

jun-rh:
> 

Ah, I see. You are right, let's change this.

jun-rh (approved):

greyerof (approved):

PR #520, opened 2021-12-13
Add diagnostic package unit tests

sebrandon1:
I'm not sure why there were so many global variables only used in one location so I moved them to be declared locally to the functions using them.

Also, I'm not positive as to the best way to unit test the `Reel` functions yet that depend on reaching out to OpenShift via Reel so I didn't try and tackle them here.

### DIFF
--- a/test-network-function/suite_test.go
+++ b/test-network-function/suite_test.go
@@ -180,12 +180,13 @@ func TestTest(t *testing.T) {
 
 // incorporateTNFVersion adds the TNF version to the claim.
 func incorporateVersions(claimData *claim.Claim) {
+	versionsOcp := diagnostic.GetVersionsOcp()
 	claimData.Versions = &claim.Versions{
 		Tnf:          gitDisplayRelease,
 		TnfGitCommit: GitCommit,
-		OcClient:     diagnostic.GetVersionsOcp().Oc,
-		Ocp:          diagnostic.GetVersionsOcp().Ocp,
-		K8s:          diagnostic.GetVersionsOcp().K8s,
+		OcClient:     versionsOcp.Oc,
+		Ocp:          versionsOcp.Ocp,
+		K8s:          versionsOcp.K8s,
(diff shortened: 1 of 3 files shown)


test-network-function/suite_test.go
```diff
@@ -243,9 +244,9 @@ func generateNodes() map[string]interface{} {
...
 	)
 	nodes := map[string]interface{}{}
-	nodes[nodeSummaryField] = diagnostic.GetNodeSummary()
-	nodes[cniPluginsField] = diagnostic.GetCniPlugins()
-	nodes[nodesHwInfo] = diagnostic.GetNodesHwInfo()
-	nodes[csiDriverInfo] = diagnostic.GetCsiDriverInfo()
+	nodes[nodeSummaryField] = make(map[string]interface{})
+	nodes[cniPluginsField] = make([]diagnostic.CniPlugin, 0)
+	nodes[nodesHwInfo] = diagnostic.NodesHwInfo{}
+	nodes[csiDriverInfo] = diagnostic.NewCsiDriverInfo()
```
javierpena, line 250: I think this is not going to work. Those fields in the "nodes" map are used to report the k8s cluster information in the claim.json file, and with this we are just creating empty structs. Thus, we're always getting null information from the test suite.

Something similar is happening in test-network-function/diagnostic/suite.go, where the global variables are replaced by local ones in some cases, so the recorded information is never propagated outside the test itself.

javierpena (changes requested):

test-network-function/suite_test.go
sebrandon1: Looking at this, I agree.  I'll mark this work-in-progress.

The main thing that I don't like about this section of code is all of the global variables populated by what is pretty much an `init` function.

sebrandon1:
Closing in favor of: https://github.com/test-network-function/test-network-function/pull/528

PR #521, opened 2021-12-13
run networking tests once per pod

hamadise:
Quick fix to avoid unnecessarily repeating ping tests among the same pods. Will refactor the pod/container data structure and auto discovery post 3.2 release.


jun-rh (approved):

greyerof (approved):

test-network-function/networking/suite.go
```diff
@@ -196,13 +202,20 @@ func runNetworkingTests(netsUnderTest map[string]netTestContext, count int) {
...
 				if _, ok := env.ContainersToExcludeFromConnectivityTests[cut.ContainerIdentifier]; ok {
 					tnf.ClaimFilePrintf("Skipping container %s because it is excluded from connectivity tests (default interface)", cut.PodName)
 					continue
 				}
+				if _, ok := m[cut.PodName]; ok {
+					// test once per podname since all containers share the
+					// same network
+					continue
+				}
+				m[cut.PodName] = true
```
javierpena, line 218: Should we do the same in testDefaultNetworkConnectivity()?

test-network-function/networking/suite.go
```diff
@@ -177,7 +177,13 @@ func runNetworkingTests(netsUnderTest map[string]netTestContext, count int) {
 		if len(netUnderTest.destTargets) == 0 {
 			ginkgo.Skip(fmt.Sprintf("There are no containers to ping for network %s. A minimum of 2 containers is needed to run a ping test (a source and a destination) Skipping test", netName))
 		}
+		m := make(map[string]bool)
 		for _, aDestIP := range netUnderTest.destTargets {
+			podName := aDestIP.containerIdentifier.PodName
+			if _, ok := m[podName]; ok {
+				continue
+			}
+			m[podName] = true
```
javierpena, line 186: Is this required, if we have made the check earlier in testDefaultNetworkConnectivity)() ?

test-network-function/networking/suite.go
hamadise: updated.

test-network-function/networking/suite.go
hamadise: updated

PR #524, opened 2021-12-14
Affiliated-certification TS: ensure all containers/operators are tested.

greyerof:
Avoid stopping the TCs when some of the containers/operators fails the
certification check. I had to replace the gomega.Eventually() for a custom 
wait-like function similar to the lifecycle's TS one.


test-network-function/certification/suite.go
```diff
@@ -54,6 +54,39 @@ var _ = ginkgo.Describe(common.AffiliatedCertTestKey, func() {
 	}
 })
 
+// getContainerCertificationRequestFunction returns function that will try to get the certification status (CCP) for a container.
+func getContainerCertificationRequestFunction(repository, containerName string) func() bool {
+	return func() bool {
+		return certAPIClient.IsContainerCertified(repository, containerName)
```
javierpena, line 60: I think we have an issue with this call, and IsOperatorCertified() too. Both calls do not differentiate the case where the container is not certified and the case where the call returned an error. So, for a non-certified container, we'll be repeating the same call over and over until the time out kicks in.

Looking at the API calls, I think it might be easier to move the busy wait down the stack, either to this function (with some changes to check for an error in the call) or even to getRequest() in internal/api/api.go, checking for specific HTTP error codes and retrying when needed.

greyerof: You're right. Can we address that in a different PR, please? With this one I just wanted to avoid stopping on the first failure.

javierpena: Ok, I didn't realize that we already had the issue before this PR, too. Let's then move on with this PR, and fix the other issue as a follow-up.

jun-rh (approved):

javierpena (approved):

sebrandon1 (approved):

PR #525, opened 2021-12-14
Stop passing REDHAT_RHEL_REGISTRY in run-container.sh

javierpena:
We stopped using that environment variable with [1], so it is no longer
needed.

[1] - https://github.com/test-network-function/cnf-certification-test-partner/pull/67


jun-rh (approved):

script/run-container.sh
```diff
@@ -108,7 +108,6 @@ ${TNF_CONTAINER_CLIENT} run --rm $DNS_ARG \
 	-e TNF_PARTNER_REPO=$TNF_PARTNER_REPO \
 	-e TNF_DEPLOYMENT_TIMEOUT=$TNF_DEPLOYMENT_TIMEOUT \
 	-e TNF_OC_DEBUG_IMAGE_ID=$TNF_OC_DEBUG_IMAGE_ID \
```
jun-rh, line 110: this can be removed too

sebrandon1 (approved):

PR #529, opened 2021-12-15
Remove TNF_OC_DEBUG_IMAGE_ID env var

javierpena:
We no longer need to specify the TNF_OC_DEBUG_IMAGE_ID environment
variable, since we are using the new debug pod image. This commit
removes all usage of the env var, and updates documentation to reflect
the current situation.


sebrandon1 (approved):

pkg/tnf/handlers/common/command.go
```diff
@@ -2,29 +2,19 @@ package common
...
-var OcDebugImageID string
-
 const (
 	ocCommand    = "oc"
 	debugCommand = "debug"
-	imageArg     = "--image"
 )
 
 // GetOcDebugCommand returns the command base for any test handler that uses the oc debug command
 func GetOcDebugCommand() string {
```
jun-rh, line 11: this whole file can probably be removed. it was introduced solely for customized image id

javierpena: Good idea, I'll check it out.

pkg/tnf/handlers/readremotefile/readremotefile.go
```diff
@@ -43,7 +42,7 @@ func NewReadRemoteFile(timeout time.Duration, nodeName, filePath string) *ReadRe
 		timeout: timeout,
 		result:  tnf.ERROR,
 		args: []string{
-			"echo", "\"cat /host" + filePath + "\"", "|", common.GetOcDebugCommand(), "-q", "node/" + nodeName,
+			"echo", "\"cat /host" + filePath + "\"", "|", "oc", "debug", "-q", "node/" + nodeName,
```
javierpena, line 45: Actually, I think this readremotefile handler could be completely removed, I can't find any place using it. However, let's do that post-3.2

jun-rh: Yes, we have quite a few handlers that should have been removed/deprecated.

jun-rh (approved):

PR #533, opened 2021-12-15
Operator TS: ensure all operators are tested.

greyerof:
Avoid stoping TC when any operator fails the check.


sebrandon1 (approved):

test-network-function/operator/suite.go
```diff
@@ -83,25 +85,32 @@ var _ = ginkgo.Describe(testSpecName, func() {
...
 	ginkgo.It(testID, func() {
+		badOperators := []configsections.Operator{}
 		for _, operatorInTest := range env.OperatorsUnderTest {
 			ginkgo.By(fmt.Sprintf("%s in namespace %s Should have a valid subscription", operatorInTest.SubscriptionName, operatorInTest.Namespace))
-			testOperatorIsInstalledViaOLM(operatorInTest.SubscriptionName, operatorInTest.Namespace)
+			values := make(map[string]interface{})
+			values["SUBSCRIPTION_NAME"] = operatorInTest.SubscriptionName
+			values["SUBSCRIPTION_NAMESPACE"] = operatorInTest.Namespace
+			tester, handlers := utils.NewGenericTesterAndValidate(relativecheckSubscriptionTestPath, relativeSchemaPath, values)
+			context := common.GetContext()
```
javierpena, line 95: Maybe we could optimize this a bit by generating the context once, out of the loop? I'm fine with doing it as a follow-up review, though.

javierpena (approved):

test-network-function/operator/suite.go
greyerof: Yes, we'll review this issue soon for all TCs+TSs.

hamadise (approved):

PR #536, opened 2021-12-16
Networking TS: ensure all networks/namespaces are tested.

greyerof:
- Avoid stopping the default/multus network ping test when some network
  fails the check.
- Avoid stopping the nodePort TC when some namespace fails the check.


test-network-function/networking/suite.go
```diff
@@ -167,16 +167,19 @@ func processContainerIpsPerNet(containerID *configsections.ContainerIdentifier,
 
 // runNetworkingTests takes a map netTestContext, e.g. one context per network attachment
 // and runs pings test with it
-func runNetworkingTests(netsUnderTest map[string]netTestContext, count int) {
+func runNetworkingTests(netsUnderTest map[string]netTestContext, count int) (badNets map[string][]string) {
```
sebrandon1, line 170: Instead of declaring `badNets` in the function's return value, on 177 just use `badNets := make(map[string][]string)`.

javierpena (approved):

test-network-function/networking/suite.go
greyerof: Done, by I used badNets := map[string][]string{} instead.

sebrandon1 (approved):

test-network-function/networking/suite.go
```diff
@@ -190,14 +193,18 @@ func runNetworkingTests(netsUnderTest map[string]netTestContext, count int) {
...
-				netUnderTest.testerSource.containerIdentifier.ContainerUID,
-				netUnderTest.testerSource.containerIdentifier.ContainerRuntime,
-				aDestIP.ip,
-				count)
+			testPass := testPing(netUnderTest.testerContainerNodeOc, netUnderTest.testerSource.containerIdentifier, aDestIP, count)
+			if !testPass {
+				if failedDestIps, netFound := badNets[netName]; netFound {
+					badNets[netName] = append(failedDestIps, aDestIP.ip)
+				} else {
+					badNets[netName] = []string{aDestIP.ip}
```
jun-rh, line 201: If we don't have dest pod, we should probably include the source pod/ip at least?

greyerof: I already modified the claim file log trace to include both pod names and the target IP. See lines 284+288.

PR #537, opened 2021-12-16
Test that check if port is declared and listening

shimritproj:
This test checks if the port is declared and listens and returns all the required ports.

### DIFF
--- a/test-network-function/networking/suite.go
+++ b/test-network-function/networking/suite.go
@@ -18,9 +18,14 @@ package networking
 
 import (
 	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
 
 	"github.com/test-network-function/test-network-function/pkg/config"
 	"github.com/test-network-function/test-network-function/pkg/tnf/testcases"
+	"github.com/test-network-function/test-network-function/pkg/utils"
 
 	"github.com/test-network-function/test-network-function/test-network-function/common"
 	"github.com/test-network-function/test-network-function/test-network-function/identifiers"
@@ -37,8 +42,34 @@ import (
 )
 
 const (
-	defaultNumPings = 5
+	CommandPortDeclared = "oc get pod %s -n %s -o json  | jq -r '.spec.containers[%d].ports'"
+	CommandPortListen   = "ss -tulwnH"
+	defaultNumPings     = 5
+	ocCommandTimeOut    = time.Second * 10
+	indexProtocolName   = 0
+	indexPort           = 4
 )
+
+var (
+	expectersVerboseModeEnabled = false
+	PortDeclaredAndListen       []PortDeclared
+	nodeListen                  PortListen
+	protocolName                = ""
+	port                        = ""
+	name                        = ""
+	flag                        = 0
+)
+
+type PortDeclared struct {
+	containerPort int
+	name          string
+	protocol      string
+}
+
+type PortListen struct {
+	port     int
+	protocol string
+}
 
 //
 // All actual test code belongs below here.  Utilities belong above.
@@ -68,6 +99,9 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
 		})
 		ginkgo.Context("Should not have type of nodePort", func() {
 			testNodePort(env)
+		})
+		ginkgo.Context("Should not have type of listen port and declared port", func() {
+			testListenAndDeclared(env)
 		})
 	}
 })
@@ -163,4 +197,111 @@ func testNodePort(env *config.TestEnvironment) {
 			test.RunAndValidate()
 		}
 	})
+}
+
+func getDeclaredPortList(command string, containerNum int, podName, podNamespace string) []PortDeclared {
+	var result []PortDeclared
+	ocCommandToExecute := fmt.Sprintf(command, podName, podNamespace, containerNum)
+	res := utils.ExecuteCommand(ocCommandToExecute, ocCommandTimeOut, interactive.GetContext(expectersVerboseModeEnabled), func() {
+		log.Error("can't run command: ", command)
+	})
+	x := strings.Split(res, "\n")
+	for _, i := range x {
+		fmt.Println(i)
+		if strings.Contains(i, "containerPort") {
+			s := strings.Split(i, ":")
+			s = strings.Split(s[1], ",")
+			port = s[0]
+		}
+		if strings.Contains(i, "name") {
+			s := strings.Split(i, ":")
+			s = strings.Split(s[1], ",")
+			name = s[0]
+		}
+		if strings.Contains(i, "protocol") {
+			s := strings.Split(i, ":")
+			protocolName = s[1]
+			flag = 1
+		}
+		if flag < 1 {
+			continue
+		}
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		noderes := PortDeclared{containerPort: p, name: name, protocol: protocolName}
+		result = append(result, noderes)
+		flag = 0
+	}
+	return result
+}
+
+func getListeningPortList(ocCommand string) []PortListen {
+	res := utils.ExecuteCommand(ocCommand, ocCommandTimeOut, interactive.GetContext(expectersVerboseModeEnabled), func() {
+		log.Error("can't run command: ", ocCommand)
+	})
+	splitRes := strings.Split(res, "\n")
+	var result []PortListen
+	protocolName := ""
+	for _, line := range splitRes {
+		fields := strings.Fields(line)
+		protocolName = fields[indexProtocolName]
+		s := strings.Split(fields[indexPort], ":")
+		p, _ := strconv.Atoi(s[1])
+		nodeListen = PortListen{port: p, protocol: protocolName}
+		result = append(result, nodeListen)
+	}
+	return result
+}
+
+func compareList(declaredPortList []PortDeclared, listeningPortList []PortListen) []PortDeclared {
+	var result []PortDeclared
+	var temp PortDeclared
+	indexElement1 := 0
+	indexElement2 := 0
+
+	for indexElement1 < len(declaredPortList) && indexElement2 < len(listeningPortList) {
+		if declaredPortList[indexElement1].containerPort == listeningPortList[indexElement2].port {
+			temp = declaredPortList[indexElement1]
+			result = append(result, temp)
+			indexElement1++
+			indexElement2++
+			continue
+		}
+		if declaredPortList[indexElement1].containerPort < listeningPortList[indexElement2].port {
+			indexElement1++
+			continue
+		}
+		indexElement2++
+	}
+	return result
+}
+
+func testListenAndDeclared(env *config.TestEnvironment) []PortDeclared {
+	var declaredPort []PortDeclared
+	var listeningPort []PortListen
+	var PortDeclaredAndListen []PortDeclared
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			ContainerCount := podUnderTest.ContainerCount
+			for i := 0; i < ContainerCount; i++ {
+				podName := podUnderTest.Name
+				podNamespace := podUnderTest.Namespace
+				temp := getDeclaredPortList(CommandPortDeclared, i, podName, podNamespace)
+				declaredPort = append(declaredPort, temp...)
+			}
+			temp := getListeningPortList(CommandPortListen)
+			listeningPort = append(listeningPort, temp...)
+			// sorting the lists
+			sort.SliceStable(declaredPort, func(i, j int) bool {
+				return declaredPort[i].containerPort < declaredPort[j].containerPort
+			})
+			sort.SliceStable(listeningPort, func(i, j int) bool {
+				return listeningPort[i].port < listeningPort[j].port
+			})
+			// compare between declaredPort,listeningPort and return the common.
+			res := compareList(declaredPort, listeningPort)
+			PortDeclaredAndListen = append(PortDeclaredAndListen, res...)
+		}
+	})
+	return PortDeclaredAndListen
 }


test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
...
+
+func testListenAndDeclared(env *config.TestEnvironment) []PortDeclared {
+	var declaredPort []PortDeclared
+	var listeningPort []PortListen
+	var PortDeclaredAndListen []PortDeclared
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			ContainerCount := podUnderTest.ContainerCount
+			for i := 0; i < ContainerCount; i++ {
```
sebrandon1, line 396:
```suggestion
			for i := 0; i < podUnderTest.ContainerCount; i++ {
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
...
+				podName := podUnderTest.Name
+				podNamespace := podUnderTest.Namespace
+				temp := getDeclaredPortList(CommandPortDeclared, i, podName, podNamespace)
```
sebrandon1, line 399:
```suggestion
				temp := getDeclaredPortList(CommandPortDeclared, i, podUnderTest.Name, podUnderTest.Namespace)
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
...
+func compareList(declaredPortList []PortDeclared, listeningPortList []PortListen) []PortDeclared {
+	var result []PortDeclared
+	var temp PortDeclared
+	indexElement1 := 0
+	indexElement2 := 0
+
+	for indexElement1 < len(declaredPortList) && indexElement2 < len(listeningPortList) {
+		if declaredPortList[indexElement1].containerPort == listeningPortList[indexElement2].port {
+			temp = declaredPortList[indexElement1]
+			result = append(result, temp)
```
sebrandon1, line 374:
```suggestion
			result = append(result, declaredPortList[indexElement1])
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
...
+		p, _ := strconv.Atoi(s[1])
+		nodeListen = PortListen{port: p, protocol: protocolName}
+		result = append(result, nodeListen)
+	}
+	return result
+}
+
+func compareList(declaredPortList []PortDeclared, listeningPortList []PortListen) []PortDeclared {
+	var result []PortDeclared
+	var temp PortDeclared
```
sebrandon1, line 367:
```suggestion
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
...
+		}
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		noderes := PortDeclared{containerPort: p, name: name, protocol: protocolName}
+		result = append(result, noderes)
+		flag = 0
+	}
+	return result
+}
+
+func getListeningPortList(ocCommand string) []PortListen {
```
sebrandon1, line 347: Might I suggest returning `([]PortListen, protocolName)`?  So we don't have have to use global variables for this.

test-network-function/networking/suite.go
```diff
@@ -39,9 +44,34 @@ import (
...
+)
+
+var (
+	expectersVerboseModeEnabled = false
+	PortDeclaredAndListen       []PortDeclared
+	nodeListen                  PortListen
+	protocolName                = ""
+	port                        = ""
+	name                        = ""
+	flag                        = 0
```
sebrandon1, line 62: I'm not sure I like all of these global variables.  We should try to minimize their usage throughout our code in general.  Could we make the functions that find these variables just return them?

sebrandon1 (changes requested):
Left some comments.

Also, I think we need unit tests for these functions as well.

Thanks!

test-network-function/networking/suite.go
shimritproj: I agree , fixed.

test-network-function/networking/suite.go
shimritproj: I agree , fixed.

test-network-function/networking/suite.go
shimritproj: fixed.

test-network-function/networking/suite.go
shimritproj: fixed.

test-network-function/networking/suite.go
shimritproj: I didn't need to return here the variable protocol Name because this variable included in struct []PortListen.

test-network-function/networking/suite.go
shimritproj: I agree, fixed.

test-network-function/networking/suite.go
```diff
@@ -34,14 +39,40 @@ import (
...
 )
 
+type PortDeclared struct {
+	containerPort int
```
hamadise, line 66: containerPort can be named simply port.

shimritproj: I agree, fixed.

test-network-function/networking/suite.go
```diff
@@ -275,3 +309,106 @@ func testNodePort(env *config.TestEnvironment) {
...
+		protocolName = fields[indexProtocolName]
+		s := strings.Split(fields[indexPort], ":")
+		p, _ := strconv.Atoi(s[1])
+		nodeListen = PortListen{port: p, protocol: protocolName}
+		result = append(result, nodeListen)
+	}
+	return result
+}
+
+func compareList(declaredPortList []PortDeclared, listeningPortList []PortListen) []PortDeclared {
```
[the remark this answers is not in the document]
shimritproj: no, compareList function that compare and return just only the common between 2 lists (declaredPortList and listeningPortList).

test-network-function/networking/suite.go
```diff
@@ -275,3 +309,106 @@ func testNodePort(env *config.TestEnvironment) {
...
+}
+
+func getListeningPortList(ocCommand string) []PortListen {
+	res, _ := utils.ExecuteCommand(ocCommand, ocCommandTimeOut, interactive.GetContext(expectersVerboseModeEnabled))
+	splitRes := strings.Split(res, "\n")
+	var result []PortListen
+	protocolName := ""
+	for _, line := range splitRes {
+		fields := strings.Fields(line)
+		protocolName = fields[indexProtocolName]
```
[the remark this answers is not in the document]
shimritproj: I agree, fixed.

test-network-function/networking/suite.go
```diff
@@ -275,3 +309,106 @@ func testNodePort(env *config.TestEnvironment) {
...
+		if declaredPortList[indexElement1].containerPort < listeningPortList[indexElement2].port {
+			indexElement1++
+			continue
+		}
+		indexElement2++
+	}
+	return result
+}
+
+func testListenAndDeclared(env *config.TestEnvironment) []PortDeclared {
```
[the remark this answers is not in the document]
shimritproj: this function checked which port is declared and listening and returned list of the ports that declared  && listening.

test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
 		}
 	})
 }
+
+func getDeclaredPortList(command string, containerNum int, podName, podNamespace string) []PortDeclared {
```
jun-rh: I don't think the command should be an argument. This function works with only one command. It's not generic.

jun-rh: Same goes with the other function

test-network-function/networking/suite.go
```diff
@@ -275,3 +308,110 @@ func testNodePort(env *config.TestEnvironment) {
...
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		noderes := PortDeclared{containerPort: p, name: name, protocol: protocolName}
+		result = append(result, noderes)
+		flag = 0
+	}
+	return result
+}
+
+func getListeningPortList(ocCommand string) []PortListen {
+	res := utils.ExecuteCommand(ocCommand, ocCommandTimeOut, interactive.GetContext(expectersVerboseModeEnabled), func() {
```
jun-rh: executing the command locally returns the listening ports on the host where our test is running. we need the the listening port from the pod. again, look at David's ping handler implementation.

test-network-function/networking/suite.go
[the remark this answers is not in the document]
jun-rh: You, me and Michael had this discussion before. The goal is to check whether all the listen ports are declared. Simply return the list doesn't test anything as Salah pointed out.

test-network-function/networking/suite.go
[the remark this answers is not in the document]
hamadise: then call it "intersection()" . Also, it's not intersection per se as the types of inputs are not the same. The name "compareList" is wrong and not reflective of what the function is doing.

test-network-function/networking/suite.go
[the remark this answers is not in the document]
shimritproj: Right, but we need to check which port is declare and listen and finally we need to return the common, I agree that this not test.

test-network-function/networking/suite.go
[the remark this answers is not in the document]
shimritproj: I agree that this name "compareList" is not correct, I changed. @hamadise

test-network-function/networking/suite.go
shimritproj: I agree, fixed.

test-network-function/networking/suite.go
```diff
@@ -275,3 +299,101 @@ func testNodePort(env *config.TestEnvironment) {
...
+			s := strings.Split(i, ":")
+			protocolName = s[1]
+			continue
+		}
+		if !(port == "" && name == "" && protocolName == "") {
+			continue
+		}
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		noderes := PortDeclared{port: p, name: name, protocol: protocolName}
+		result = append(result, noderes)
```
sebrandon1, line 330:
```suggestion
		result = append(result, PortDeclared{port: p, name: name, protocol: protocolName})
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +299,101 @@ func testNodePort(env *config.TestEnvironment) {
...
+	var result []PortListen
+	for _, line := range lines {
+		fields := strings.Fields(line)
+		if indexProtocolName > len(fields) || indexPort > len(fields) {
+			return result
+		}
+		s := strings.Split(fields[indexPort], ":")
+		p, _ := strconv.Atoi(s[1])
+		nodeListen := PortListen{port: p, protocol: fields[indexProtocolName]}
+		result = append(result, nodeListen)
```
sebrandon1, line 347:
```suggestion
		result = append(result, PortListen{port: p, protocol: fields[indexProtocolName]})
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +299,101 @@ func testNodePort(env *config.TestEnvironment) {
...
+	var PortDeclaredAndListen []PortDeclared
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				temp := getDeclaredPortList(CommandPortDeclared, i, podUnderTest.Name, podUnderTest.Namespace)
+				declaredPort = append(declaredPort, temp...)
+			}
+			temp := getListeningPortList(CommandPortListen)
+			listeningPort = append(listeningPort, temp...)
```
sebrandon1, line 385:
```suggestion
			listeningPort = append(listeningPort, getListeningPortList(CommandPortListen)...)
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +299,101 @@ func testNodePort(env *config.TestEnvironment) {
...
+			// sorting the lists
+			sort.SliceStable(declaredPort, func(i, j int) bool {
+				return declaredPort[i].port < declaredPort[j].port
+			})
+			sort.SliceStable(listeningPort, func(i, j int) bool {
+				return listeningPort[i].port < listeningPort[j].port
+			})
+			// compare between declaredPort,listeningPort and return the common.
+			res := compareList(declaredPort, listeningPort)
+			PortDeclaredAndListen = append(PortDeclaredAndListen, res...)
```
sebrandon1, line 395:
```suggestion
			PortDeclaredAndListen = append(PortDeclaredAndListen, compareList(declaredPort, listeningPort)...)
```

test-network-function/networking/suite.go
```diff
@@ -275,3 +299,101 @@ func testNodePort(env *config.TestEnvironment) {
...
+func testListenAndDeclared(env *config.TestEnvironment) []PortDeclared {
+	var declaredPort []PortDeclared
+	var listeningPort []PortListen
+	var PortDeclaredAndListen []PortDeclared
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				temp := getDeclaredPortList(CommandPortDeclared, i, podUnderTest.Name, podUnderTest.Namespace)
+				declaredPort = append(declaredPort, temp...)
```
sebrandon1, line 382:
```suggestion
				declaredPort = append(declaredPort, getDeclaredPortList(CommandPortDeclared, i, podUnderTest.Name, podUnderTest.Namespace)...)
```

sebrandon1 (commented):
Just a few nits.

test-network-function/networking/suite.go
```diff
@@ -275,3 +300,123 @@ func testNodePort(env *config.TestEnvironment) {
...
+			for _, cut := range env.ContainersUnderTest {
+				fmt.Println(cut)
+				if cut.ContainerIdentifier.PodName == podUnderTest.Name && cut.ContainerIdentifier.Namespace == podUnderTest.Namespace {
+					x = cut
+					break
+				}
+			}
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+
+			temp := []string{utils.AddNsenterPrefix(containerPID), CommandPortListen, "-tulwnH"}
```
jun-rh, line 403: why not put arguments into the const?

test-network-function/networking/suite.go
```diff
@@ -32,16 +37,33 @@ import (
...
 	"github.com/test-network-function/test-network-function/pkg/tnf/interactive"
 	"github.com/test-network-function/test-network-function/pkg/tnf/reel"
-	"github.com/test-network-function/test-network-function/pkg/utils"
+
 	"github.com/test-network-function/test-network-function/test-network-function/results"
 )
 
 const (
-	defaultNumPings = 5
+	CommandPortDeclared = "oc get pod %s -n %s -o json  | jq -r '.spec.containers[%d].ports'"
```
jun-rh, line 48: lower case if private

test-network-function/networking/suite.go
```diff
@@ -32,16 +37,33 @@ import (
...
+	CommandPortListen   = "ss"
+	defaultNumPings     = 5
+	ocCommandTimeOut    = time.Second * 10
+	indexProtocolName   = 0
+	indexPort           = 4
 )
 
+type PortDeclared struct {
```
jun-rh, line 56: same here

test-network-function/networking/suite.go
```diff
@@ -275,3 +300,123 @@ func testNodePort(env *config.TestEnvironment) {
...
+			}
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+
+			temp := []string{utils.AddNsenterPrefix(containerPID), CommandPortListen, "-tulwnH"}
+			fmt.Println(temp)
+			temp1 := getListeningPortList(temp)
+			listeningPort = append(listeningPort, temp1...)
+
+			// sorting the lists
+			sort.SliceStable(declaredPort, func(i, j int) bool {
```
jun-rh, line 409: compare and sort should include protocol too

jun-rh: better move the sorting into the compare function and make some good unit test for it, i.e. different port/protocol combinations in different orders

test-network-function/networking/suite.go
```diff
@@ -275,3 +300,123 @@ func testNodePort(env *config.TestEnvironment) {
...
+	res := true
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				temp := getDeclaredPortList(i, podUnderTest.Name, podUnderTest.Namespace)
+				declaredPort = append(declaredPort, temp...)
+			}
+
+			nodeName := podnodename.NewPodNodeName(common.DefaultTimeout, podUnderTest.Name, podUnderTest.Namespace)
```
hamadise, line 388: the nodename information is store inside container, using a handler to get this information is an overkill, you can add another loop to get it.

test-network-function/networking/suite.go
hamadise: I would add on this, it's lot easier to compare maps than to compare ports. 
in all cases, a good unit test as Jun suggested is required.

test-network-function/networking/suite.go
```diff
@@ -275,3 +300,123 @@ func testNodePort(env *config.TestEnvironment) {
...
+	indexElement1 := 0
+	indexElement2 := 0
+
+	for indexElement1 < len(declaredPortList) && indexElement2 < len(listeningPortList) {
+		if declaredPortList[indexElement1].port == listeningPortList[indexElement2].port {
+			indexElement1++
+			indexElement2++
+			continue
+		}
+		if listeningPortList[indexElement2].port < declaredPortList[indexElement1].port {
```
hamadise, line 367: you should not assume the order, just test if an element is different.
```if listeningPortList[indexElement2].port != declaredPortList[indexElement1].port ```
Also, this is not enough,  you need to make sure they have the same length, otherwise they are not equal.
For this, only one index is needed

File: test-network-function/networking/suite.go (hunk 21 above)
Comment (shimritproj) [reply test-network-function/networking/suite.go:409]: I agree and I fixed it.

File: test-network-function/networking/suite.go (hunk 23 above)
Comment (shimritproj) [reply test-network-function/networking/suite.go:367]: I fixed it.

File: test-network-function/networking/suite.go (hunk 14 above)
Comment (shimritproj) [reply test-network-function/networking/suite.go:352]: finally I was use in map and not use in struct.

File: test-network-function/networking/suite.go [hunk 24]
```diff
@@ -275,3 +293,96 @@ func testNodePort(env *config.TestEnvironment) {
 		}
 	})
 }
+
+func getDeclaredPortList(container int, podName, podNamespace string) {
```
sebrandon1, line 297: I think that `declaredPort` should be the return value from this function versus setting a global variable.

test-network-function/networking/suite.go
```diff
@@ -275,3 +293,96 @@ func testNodePort(env *config.TestEnvironment) {
...
+		}
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		name = strings.TrimSpace(strings.ReplaceAll(name, "\"", ""))
+		protocolName = strings.TrimSpace(strings.ReplaceAll(protocolName, "\"", ""))
+		declaredPort[p] = name + " " + protocolName
+		port, name, protocolName = "", "", ""
+	}
+}
+
+func getListeningPortList(commandlisten []string, nodeOc *interactive.Context) {
```
sebrandon1, line 328: Same thing with this function.  I think it should return the actual value instead of setting the global variable.

test-network-function/networking/suite.go
```diff
@@ -275,3 +293,96 @@ func testNodePort(env *config.TestEnvironment) {
...
+		if indexprotocolname > len(fields) || indexport > len(fields) {
+			return
+		}
+		s := strings.Split(fields[indexport], ":")
+		p, _ := strconv.Atoi(s[1])
+		listeningPort[p] = fields[indexprotocolname]
+	}
+}
+
+func checkIfListenIsDeclared() bool {
```
sebrandon1, line 344: Pass in the `declaredPort` and `listeningPort` variables and compare them versus using global variables.

Also there should probably be unit tests wrapping this function.

test-network-function/networking/suite.go
shimritproj: you're right, I fixed.

test-network-function/networking/suite.go
shimritproj: I fixed.

test-network-function/networking/suite.go
shimritproj: I fixed.

test-network-function/networking/suite.go
```diff
@@ -32,14 +36,25 @@ import (
...
+	commandportlisten   = "ss -tulwnH"
+	defaultNumPings     = 5
+	ocCommandTimeOut    = time.Second * 10
+	indexprotocolname   = 0
+	indexport           = 4
+)
+
+var (
+	declaredPort  = make(map[int]string)
+	listeningPort = make(map[int]string)
```
sebrandon1, line 57: Are these still needed since you changed the functions to return the values?

shimritproj: no no , I was forget to delete this , I delete now.

test-network-function/networking/suite.go
```diff
@@ -323,3 +336,100 @@ func testNodePort(env *config.TestEnvironment) {
...
+			return false
+		}
+	}
+	return true
+}
+
+func testListenAndDeclared(env *config.TestEnvironment) {
+	declaredPort := make(map[int]string)
+	listeningPort := make(map[int]string)
+	var x *config.Container
```
javierpena, line 402: Looking at the code, it looks like the Container type is defined in the configsections package (pkg/config/configsections) instead of the config one

PR #546, opened 2021-12-20
Removing naked returns

sebrandon1:
Related to: https://issues.redhat.com/browse/CNFCERT-249

Naked returns are less explicit and less readable and to be consistent across our code we should probably stop using them.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -62,65 +62,72 @@ func (api CertAPIClient) IsOperatorCertified(org, packageName string) bool {
 	return true
 }
 
-// GetImageByID get container image data for the given container Id
-func (api CertAPIClient) GetImageByID(id string) (response string, err error) {
-	var responseData []byte
+// GetImageByID get container image data for the given container Id.  Returns (response, error).
+func (api CertAPIClient) GetImageByID(id string) (string, error) {
+	var response string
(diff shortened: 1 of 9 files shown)


internal/api/api.go
```diff
@@ -63,64 +63,70 @@ func (api CertAPIClient) IsOperatorCertified(org, packageName string) bool {
 }
 
 // GetImageByID get container image data for the given container Id
-func (api CertAPIClient) GetImageByID(id string) (response string, err error) {
-	var responseData []byte
+func (api CertAPIClient) GetImageByID(id string) (string, error) {
```
hamadise, line 66: My opinion on this is it's good to use in functions with few lines of code such as this one where the logic is not complicated and you can think of that name in signature as the declaration of variable itself. 

Is it recommended not to use this future altogether or only when the length/complexity of a function is "significant" ?

greyerof: The output variable names are actually a good way to auto-document them. So, in case they are removed, I think we should document them in the function's comment.

javierpena: I like the idea of removing naked returns, but as @greyerof mentioned it would be nice to document the function ouptut then.

sebrandon1: Sounds good, I'll add the output in the comment.

javierpena (approved):

hamadise (approved):

PR #547, opened 2021-12-21
Update CATALOG.md

jgato:
I think this would help to better reading as (classified) catalog and it is easier to mach with the 'General Suites Test' table in the main document. The "classification" also would help in a table of content.
I have also found some mismatch between some descriptions and test naming, or, some duplicated.

### DIFF
--- a/CATALOG.md
+++ b/CATALOG.md
@@ -1,13 +1,15 @@
-# test-network-function test case catalog
+# General Tests for test-network-function
 
 test-network-function contains a variety of `Test Cases`, as well as `Test Case Building Blocks`.
 * Test Cases:  Traditional JUnit testcases, which are specified internally using `Ginkgo.It`.  Test cases often utilize several Test Case Building Blocks.
 * Test Case Building Blocks:  Self-contained building blocks, which perform a small task in the context of `oc`, `ssh`, `shell`, or some other `Expecter`.
-## Test Case Catalog
 
-Test Cases are the specifications used to perform a meaningful test.  Test cases may run once, or several times against several targets.  CNF Certification includes a number of normative and informative tests to ensure CNFs follow best practices.  Here is the list of available Test Cases:
-### http://test-network-function.com/testcases/access-control/cluster-role-bindings
+## Tests Cases Catalog
 
+Test Cases are the specifications used to perform a meaningful test.  Test cases may run once, or several times against several targets.  CNF Certification includes a number of normative and informative tests to ensure CNFs follow best practices.  Here is the list of available Test Cases
+
+### access-control
+#### cluster-role-bindings
 Property|Description
 ---|---
 Version|v1.0.0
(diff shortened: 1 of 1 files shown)


greyerof (commented):
The CATALOG.md file is generated automatically using the tnf tool, which is also built as part as the "make build" process. During "make build", the catalog is rebuilt with:

`./tnf generate catalog markdown`

This tool uses templates to build the markdown entries from both identifiers.go files. The markdown templates are at cmd/tnf/generate/catalog.

jgato:
If it is automatic generated there are some issues. For example:

 * platform-alteration/sysctl-config but describing pod-recreation. Or, the url in the description is not ok.

https://github.com/test-network-function/test-network-function/blob/main/CATALOG.md?plain=1#L279-L287

Would it make sense that sections/subsections have titles in the way of 
## suite
### test 
instead of urls?

jun-rh:
#553 created for correcting the url.

jgato:
ok.
Do the urls need to be kept because they are adding some important meaning?
Something that could be also enhanced is the name of the test. It took me a while to guess the name of test if you want to skip it from a suite. How is the correct way? I managed to skip a test with:
```
./run-tnf-container.sh  -t ~/Projects-src/certification/test-network-function/test-network-function/ -o output -s lifecycle-pod-recreation -f lifecycle
```
But in the main README, it seems you have to concatenate the SUITE-TEST for the name. Like in this example
```
./run-tnf-container.sh -k ~/.kube/config -t ~/tnf/config -o ~/tnf/output -f diagnostic access-control -s access-control-host-resource-PRIVILEGED_POD
```

jun-rh:
> ok. Do the urls need to be kept because they are adding some important meaning? Something that could be also enhanced is the name of the test. It took me a while to guess the name of test if you want to skip it from a suite. How is the correct way? I managed to skip a test with:
> 
> ```
> ./run-tnf-container.sh  -t ~/Projects-src/certification/test-network-function/test-network-function/ -o output -s lifecycle-pod-recreation -f lifecycle
> ```
> 
> But in the main README, it seems you have to concatenate the SUITE-TEST for the name. Like in this example
> 
> ```
> ./run-tnf-container.sh -k ~/.kube/config -t ~/tnf/config -o ~/tnf/output -f diagnostic access-control -s access-control-host-resource-PRIVILEGED_POD
> ```

Yes, there was a grand vision behind the use of the url as test id. This repo was meant to be the library/building blocks for test sutie development. Everything under the test-network-function was meant to be in a separate test case repo. The hope was that partners and vendors would start building their own test cases/suites by invoking the test handlers from there own test case repo. To avoid test case id collision, domain name and url becomes the easiest solution as it provides a natural hierarchy and ownership.

And yes, this does make skipping and focusing with ginkgo a lot more difficult. The assumption is that all tests should be run in normal cases. Skipping or focusing is more for development or troubleshooting. So we came up with this short id which is the test category name plus the real test name. They can be found in the test result if you need to skip or focus certain test after a full run.

In summary, the fully qualified url id is the id for the (global) catalog. The short version is local to this specific test suite (CNF certification generic test). The reason why the image is called tnf is because it's the tnf library/framework image. For now, it contains one generic test suite only. In theory, it could include any other suite built on top of the same framework.

jgato:
Hi @jun-rh 
I am working on that "new" way of printing the catalog 

https://github.com/jgato/test-network-function/tree/%231_Enhance_Catalog_Browsing

I still have some minor issues to polish and I will try the merge request, if you find this interesting.

PR #550, opened 2021-12-21
update statefulset scaling identifier

aabughosh:
update statefulset scaling identifier

### DIFF
--- a/test-network-function/identifiers/identifiers.go
+++ b/test-network-function/identifiers/identifiers.go
@@ -213,6 +213,11 @@ var (
 		Url:     formTestURL(common.LifecycleTestKey, "scaling"),
 		Version: versionOne,
 	}
+	// TestStateFulScalingIdentifier ensures deployment scale in/out operations work correctly.
+	TestStateFulScalingIdentifier = claim.Identifier{
+		Url:     formTestURL(common.LifecycleTestKey, "stateful-scaling"),
+		Version: versionOne,
+	}
 	// TestIsRedHatReleaseIdentifier ensures platform is defined
 	TestIsRedHatReleaseIdentifier = claim.Identifier{
 		Url:     formTestURL(common.PlatformAlterationTestKey, "isredhat-release"),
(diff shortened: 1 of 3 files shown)


test-network-function/identifiers/identifiers.go
```diff
@@ -213,6 +213,11 @@ var (
 		Url:     formTestURL(common.LifecycleTestKey, "scaling"),
 		Version: versionOne,
 	}
+	// TestStateFulScalingIdentifier ensures deployment scale in/out operations work correctly.
+	TestStateFulScalingIdentifier = claim.Identifier{
+		Url:     formTestURL(common.LifecycleTestKey, "stateful-scaling"),
```
jun-rh, line 218: stateful -> statefulset

test-network-function/identifiers/identifiers.go
```diff
@@ -213,6 +213,11 @@ var (
 		Url:     formTestURL(common.LifecycleTestKey, "scaling"),
```
jun-rh, line 213: to be consistent, rename this too

test-network-function/identifiers/identifiers.go
```diff
@@ -572,16 +577,26 @@ the changes for you.`,
...
 			scale-in oc command for (N-1) replicas. Lastly, it executes the scale-out oc command, restoring the original replicaCount of the deployment/s.`),
 		Remediation:           `Make sure CNF deployments/replica sets can scale in/out successfully.`,
 		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 6.2",
 	},
+	TestStateFulSetScalingIdentifier: {
+		Identifier: TestStateFulSetScalingIdentifier,
+		Type:       normativeResult,
+		Description: formDescription(TestStateFulSetScalingIdentifier,
+			`tests that CNF deployments support scale in/out operations. 
+			First, The test starts getting the current replicaCount (N) of the statefulset/s with the Pod Under Test. Then, it executes the
```
jun-rh, line 595: @greyerof  is the `with the Pod Under Test` part out of date? is it there because we used to find deployments through the pod list as part of the test? also, any suggestion on how to update the steps to reflect the HPA case?

greyerof: No, I think it's still a valid statement. Both target deployments and statefulsets are autodiscovered using the same targetPodLabels as we use for discovering the PUTs. We just search those labels on the deployment/statefulset's .spec.template.metadata.labels.

javierpena (approved):
Nice, this also improves readability of the HTML output, where one of the tests was usually skipped but collapsed with the other.

jun-rh (approved):

PR #554, opened 2021-12-22
Replace usage of ifconfig with ip

javierpena:
The ifconfig utility is deprecated in RHEL [1] and could be removed in
future releases. When gathering network information from nodes, we
should use "ip".

This commit changes the diagnostic-nodes-hw-info test to use "ip a"
instead of ifconfig, and adjusts output parsing to match the
differences.

[1] - https://access.redhat.com/articles/1179833


hamadise (approved):

javierpena:
This is the new ip output, taken from the CI job artifacts:

image

sebrandon1 (approved):

PR #559, opened 2022-01-04
Add tests for csv_info.go

sebrandon1:
Adding some more coverage for the autodiscover/csv_info functions.

### DIFF
--- a/pkg/config/autodiscover/csv_info_test.go
+++ b/pkg/config/autodiscover/csv_info_test.go
@@ -17,6 +17,7 @@
 package autodiscover
 
 import (
+	"errors"
 	"log"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


pkg/config/autodiscover/csv_info_test.go
```diff
@@ -57,3 +58,106 @@ func TestCSVGetAnnotationValue(t *testing.T) {
...
+		expectedString string
+	}{
+		{
+			expectedString: "error (this is an error) attempting to unmarshal value of annotation 'testKey' on CSV 'testnamespace/testname'",
+		},
+	}
+
+	for _, tc := range testCases {
+		csvr := CSVResource{
+			Metadata: struct {
```
greyerof, line 73: This is not important, but I'm curious: is it necessary to re-declare the whole struct types for the initialization? I'd just do this:
```
csvr := CSVResource{}
csvr.Name = "testname"
csvr.Namespace = "testnamespace"
```

pkg/config/autodiscover/csv_info_test.go
```diff
@@ -57,3 +58,106 @@ func TestCSVGetAnnotationValue(t *testing.T) {
...
+		origFunc := executeOcGetAllCommand
+		executeOcGetCommand = func(resourceType, labelQuery, namespace string) string {
+			output, err := os.ReadFile(path.Join(filePath, tc.filename))
+			assert.Nil(t, err)
+			return string(output)
+		}
+
+		outputList, err := GetCSVsByLabelByNamespace(tc.label, tc.value, "testnamespace")
+		assert.Nil(t, err)
+		if len(outputList.Items) > 0 {
```
greyerof, line 157: I believe that the number of parsed Items should also be checked with something like:
`assert.Equal(t, len(outputList.Items), tc.ExpectedItemsCount)`

pkg/config/autodiscover/csv_info_test.go
sebrandon1: Yeah that was an interesting auto-fill by `gofmt`.  Not sure why it brought all of that in.  I can double check if removing it causes any issues.

pkg/config/autodiscover/csv_info_test.go
sebrandon1: Good idea, I'll add an expected items count.

pkg/config/autodiscover/csv_info_test.go
sebrandon1: Nope, it doesn't like it when I remove it.

pkg/config/autodiscover/csv_info_test.go
sebrandon1: Done.

jun-rh (approved):

greyerof (approved):

PR #560, opened 2022-01-04
Make the networking tests run once per pod instead of once per container

edcdavid:
Stores the context of one container in the pod object to get access to the networking namespace of the pod (since all container share the same net namespace). 
The networking tests are now ran per pod instead of per container

### DIFF
--- a/pkg/config/autodiscover/autodiscover_targets.go
+++ b/pkg/config/autodiscover/autodiscover_targets.go
@@ -57,7 +57,7 @@ func FindTestTarget(labels []configsections.Label, target *configsections.TestTa
 			for _, pod := range pods.Items {
 				if ns[pod.Metadata.Namespace] {
 					target.PodsUnderTest = append(target.PodsUnderTest, buildPodUnderTest(pod))
-					target.ContainerConfigList = append(target.ContainerConfigList, buildContainersFromPodResource(pod)...)
+					target.ContainerList = append(target.ContainerList, buildContainers(pod)...)
 				} else {
 					target.NonValidPods = append(target.NonValidPods, buildPodUnderTest(pod))
 				}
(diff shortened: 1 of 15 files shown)


pkg/config/autodiscover/autodiscover_targets.go
```diff
@@ -175,8 +175,10 @@ func FindTestPodSetsByLabel(targetLabels []configsections.Label, target *configs
 }
 
 // buildPodUnderTest builds a single `configsections.Pod` from a PodResource
-func buildPodUnderTest(pr *PodResource) (podUnderTest configsections.Pod) {
+func buildPodUnderTest(pr *PodResource) (podUnderTest *configsections.Pod) {
 	var err error
+	var pod configsections.Pod
+	podUnderTest = &pod
```
sebrandon1, line 181:
```suggestion
	podUnderTest := &configsections.Pod{}
```

pkg/config/configsections/container.go
```diff
@@ -18,8 +18,82 @@ package configsections
...
+			case err := <-outCh:
+				log.Fatalf("OC session to container %s/%s is broken due to: %v, aborting the test run", oc.GetPodName(), oc.GetPodContainerName(), err)
+			case <-oc.GetDoneChannel():
+				log.Debugf("stop watching the session with container %s/%s", oc.GetPodName(), oc.GetPodContainerName())
+			}
+		}()
+		ocChan <- oc
+	}()
+
+	containerOc = <-ocChan
```
sebrandon1, line 76: Why do the above section if you are just blocking waiting for the interactive session to come back?

sebrandon1 (commented):
Just a couple of comments.

pkg/config/configsections/container.go
edcdavid: Fyi, I did not modify this code in this PR, just moved the code to resolve dependencies. I believe the reason for this section is to return the "oc" context right away and to gracefully handles normal termination and error cases. The interactive session is long lived and multiples command are ran on it. This section monitors the status in the background

sebrandon1: Yeah I saw that it was moved.  Interesting function I'll have to dig into it some more.

sebrandon1 (approved):

jun-rh:
General comments:

- Why not add the whole list of containers to the pod structure?
- Doesn't it make more sense to move the networking related field to the pod structure?

jun-rh (approved):

PR #561, opened 2022-01-05
#1 enhance catalog browsing

jgato:
I have done some modifications to create a more browseable catalog as discussed in https://github.com/test-network-function/test-network-function/pull/547

This is the visual result:

image

It should help to see/understand the list of available tests and how these are classified by test units. The current way of showing the catalog shows a list of urls that makes more difficult to follow when you look for an specific test to run or to skip.
How it was previously
image

I had to modify the two identifiers packages, in order provide/export the basedomain used for the urls or tests. I am not sure if this is ok. This is used by the new code to extract suite and test names from the url. Which is the base of the new printing.
So what it basically does it to get the two kind of urls:
 http://test-network-function.com/testcases/SuiteName/TestName
 http://test-network-function.com/tests//TestName
To extract SuiteName-TestName or just TestName for the new printing. The other information is show as it was.


cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,119 @@ func emitTextFromFile(filename string) error {
...
+// }
+func createPrintableCatalogFromIdentifiers(keys []claim.Identifier) map[string][]CatalogElement {
+	base_domain := identifiers.TestIDBaseDomain + "/"
+
+	catalog := make(map[string][]CatalogElement)
+	// we need the list of suite's names
+	var element CatalogElement
+
+	for _, i := range keys {
+		suite_test := getSuiteAndTestFromUrl(i.Url, base_domain)
```
jun-rh, line 169: This is similar to XformToGinkgoItIdentifier(). Why not use that or at least keep the base domain related logic inside the identifiers package (instead of exporting the actual domain)?

jgato: I fully agree, it would be nice to export from identifiers packages this kind of functionality. The problem with XformToGinkgoItIdentifier() is that only returns the TestName, but you cannot get Suite Name. So you cannot show the catalogue classified in the way of
SuiteName
|
|- TestName
|- TestNAme
SuiteName
|
|- Testname
we would need something opossite to formTestURL:

https://github.com/test-network-function/test-network-function/blob/main/test-network-function/identifiers/identifiers.go#L55-L58


But the other way around. So from an url you can de-compose and get the suite and test names.

jgato: @jun-rh  do you think we could add these methods to the identifier's packages?
 * pkg/tnf/identifiers/identifiers.go  has no function that allows to get the TestName from the identifier url. Here there is no SuitNAme, or SuitName is just Test Case Building Block
 * test-network-functions/identifiers/identifiers.go has the XformToGinkgoItIdentifier that we can use to get the TestName, but we need other similar to get the SuiteName.
 Do you agree to add these changes? Can we use this same pull request?

jgato: Done it. Now the functions that helps you get SuiteName and TestName from identifiers.urls moved to each corresponding package.
A little bit tricky because each package works differently about managing Identifiers. 
I have seen one of the packages has some tests. So I have added a new one for the new added function.

jun-rh:
FYI: @javierpena is adding label based test case filtering in #571. Take a look and see if it would change how we want to present the catalog.

jgato:
I think it does not change the way the catalogue is presented. But, it would be a good opportunity to point in the catalogue to the id/word to skip the test. 
Now it shows:


Property|Description
---|---
Test Name|container-shutdown
Url|http://test-network-function.com/testcases/lifecycle/container-shutdown
Version|v1.0.0

And it could be added: 


Property|Description
---|---
Test Name|container-shutdown
ID to skip | lifecycle-container-shutdown
Url|http://test-network-function.com/testcases/lifecycle/container-shutdown
Version|v1.0.0

And in the readme points to the catalogue to the get ID to Skip, in the case you want to skip some tests. Does this affects also to the Test Cases Building Block? 

"ID To skip" or other meaning attribute.

javierpena:
> And in the readme points to the catalogue to the get ID to Skip, in the case you want to skip some tests. Does this affects also to the Test Cases Building Block?
> 
> "ID To skip" or other meaning attribute.

Maybe just "Test ID"? We don't have that in the catalog, so it could be a good attribute name.

jgato:
> > And in the readme points to the catalogue to the get ID to Skip, in the case you want to skip some tests. Does this affects also to the Test Cases Building Block?
> > "ID To skip" or other meaning attribute.
> 
> Maybe just "Test ID"? We don't have that in the catalog, so it could be a good attribute name.

@jun-rh  which is be best option? Maybe first we go with this PR.When it is ok, we can make the modifications for adding Test ID, which will much less code  and easier to review.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -157,19 +254,28 @@ func outputTestCaseBuildingBlocks() {
...
-		fmt.Println("Property|Description")
-		fmt.Println("---|---")
-		fmt.Fprintf(os.Stdout, "Version|%s\n", identifier.Catalog[k].Identifier.SemanticVersion)
-		fmt.Fprintf(os.Stdout, "Description|%s\n", identifier.Catalog[k].Description)
-		fmt.Fprintf(os.Stdout, "Result Type|%s\n", identifier.Catalog[k].Type)
-		fmt.Fprintf(os.Stdout, "Intrusive|%t\n", identifier.Catalog[k].IntrusionSettings.ModifiesSystem)
-		fmt.Fprintf(os.Stdout, "Modifications Persist After Test|%t\n", identifier.Catalog[k].IntrusionSettings.ModificationIsPersistent)
-		fmt.Fprintf(os.Stdout, "Runtime Binaries Required|%s\n", cmdJoin(identifier.Catalog[k].BinaryDependencies, ", "))
-		fmt.Println()
+	catalog := createPrintableCatalogFromUrls(keys)
```
jun-rh, line 257: This section is for the handlers. They are meant be used as go library when writing test cases. So I don't see a need of updating this the same way as the test cases section.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+		suiteTest := identifiers.GetSuiteAndTestFromIdentifier(i)
+		if suiteTest == nil {
+			fmt.Fprintf(os.Stderr, "Identifier Url not valid\n")
+			return nil
+		}
+		suiteName := suiteTest[0]
+		testName := suiteTest[1]
+		element.testName = testName
+		element.identifier = i
+		catalog[suiteName] = append(catalog[suiteName], element)
```
sebrandon1, line 147:
```suggestion
		catalog[suiteName] = append(catalog[suiteName], catalogElement{
		    testName: testName,
		    identifier: i
		})
```

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+// and takes urls like http://test-network-function.com/testcases/SuiteName/TestName
+// to build a more structured catalogue
+// {
+//     suiteNameA: [testName [identifiers], testName2 [identifiers] ]
+//     suiteNameB: [testName3 [identifiers], testName4 [identifiers] ]
+// }
+func createPrintableCatalogFromIdentifiers(keys []claim.Identifier) map[string][]catalogElement {
+	catalog := make(map[string][]catalogElement)
+	// we need the list of suite's names
+	var element catalogElement
```
sebrandon1, line 135: Don't need to declare `element` here, just append to the map like my below suggestion.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+
+// it turns a list of urls http://test-network-function.com/tests/TestName
+// to build a more structured catalogue
+// {
+//     0: [testName [identifiers], testName2 [identifiers] ]
+//     1: [testName3 [identifiers], testName4 [identifiers] ]
+// }
+func createPrintableCatalogFromUrls(urls []string) map[int][]catalogElement {
+	catalog := make(map[int][]catalogElement)
+	var element catalogElement
```
sebrandon1, line 160: Same thing here, `element` doesn't need to be declared here.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+	var c = 0
```
sebrandon1, line 161: No need for the `c` counter here, just use the index in the for loop instead of `_`.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+	for _, url := range urls {
+		testName := identifier.XformToGinkgoItIdentifier(identifier.Identifier{URL: url})
+		if testName == "" {
+			return nil
+		}
+		element.testName = testName
+		element.identifier = claim.Identifier{Url: url}
+		catalog[c] = append(catalog[c], element)
```
sebrandon1, line 169:
```suggestion
		element.testName = testName
		element.identifier = claim.Identifier{Url: url}
		catalog[c] = append(catalog[c], catalogElement{
		   testName: testName,
		   identifier: claim.Identifier{
		      Url: url;
		   }
		})
```

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+		c++
```
sebrandon1, line 170: No need for `c` counter here, just use the index of the loop.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+	return catalog
+}
+
+func getSuitesFromIdentifiers(keys []claim.Identifier) []string {
+	var suites []string
+
+	for _, i := range keys {
+		suiteTest := identifiers.GetSuiteAndTestFromIdentifier(i)
+		suiteName := suiteTest[0]
+		suites = append(suites, suiteName)
```
sebrandon1, line 181:
```suggestion
		suites = append(suites, identifiers.GetSuiteAndTestFromIdentifier(i)[0])
```

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,85 @@ func emitTextFromFile(filename string) error {
...
+	uniqMap := make(map[string]struct{})
+	for _, v := range slice {
+		uniqMap[v] = struct{}{}
+	}
+
+	// turn the map keys into a slice
+	uniqSlice := make([]string, 0, len(uniqMap))
+	for v := range uniqMap {
+		uniqSlice = append(uniqSlice, v)
+		fmt.Println(v)
```
sebrandon1, line 198: Is this a debug statement?

pkg/tnf/identifier/identifier_test.go
```diff
@@ -83,3 +83,43 @@ func TestIdentifier_UnmarshalJSON(t *testing.T) {
...
+		{
+			URL:      "http://test-network-function.es/functional-tests" + "/whatever",
+			testName: "whatever",
+		},
+	}
+
+	for _, test := range testsURLOk {
+		id := identifier.Identifier{URL: test.URL, SemanticVersion: ""}
+		testName := identifier.XformToGinkgoItIdentifier(id)
+		assert.Equal(t, testName, test.testName)
```
sebrandon1, line 117:
```suggestion
		testName := 
		assert.Equal(t, test.testName, identifier.XformToGinkgoItIdentifier(id))
```

test-network-function/identifiers/identifiers.go
```diff
@@ -259,6 +261,24 @@ func XformToGinkgoItIdentifierExtended(identifier claim.Identifier, extra string
...
+func GetSuiteAndTestFromIdentifier(identifier claim.Identifier) []string {
+	result := strings.Split(identifier.Url, url+"/")
+	const SPLITN = 2
+	// len 2, the baseDomain can appear only once in the url
+	// so it returns what you have previous and before basedomain
+	if len(result) != SPLITN {
+		return nil
+	}
+	suiteTest := strings.Split(result[1], "/")
+	return suiteTest
```
sebrandon1, line 279:
```suggestion
	return strings.Split(result[1], "/")
```

sebrandon1 (changes requested):
Couple of suggestions.

cmd/tnf/generate/catalog/catalog.go
jgato: yes, it was a debug statement. No longer need it

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -116,6 +121,81 @@ func emitTextFromFile(filename string) error {
 	return nil
 }
 
+// it turns a list of identifiers
```
javierpena, line 124: For autodoc purposes, we should replace "it" with the function name, in this case "createPrintableCatalogFromIdentifiers". I see a couple more instances of this pattern in the PR.

cmd/tnf/generate/catalog/catalog.go
jgato: done it

cmd/tnf/generate/catalog/catalog.go
jgato: done it

cmd/tnf/generate/catalog/catalog.go
jgato: done it

cmd/tnf/generate/catalog/catalog.go
jgato: you are right I have tried to improve the doc of the two main functions I created.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -157,19 +256,28 @@ func outputTestCaseBuildingBlocks() {
...
-		fmt.Fprintf(os.Stdout, "Intrusive|%t\n", identifier.Catalog[k].IntrusionSettings.ModifiesSystem)
-		fmt.Fprintf(os.Stdout, "Modifications Persist After Test|%t\n", identifier.Catalog[k].IntrusionSettings.ModificationIsPersistent)
-		fmt.Fprintf(os.Stdout, "Runtime Binaries Required|%s\n", cmdJoin(identifier.Catalog[k].BinaryDependencies, ", "))
-		fmt.Println()
+	catalog := createPrintableCatalogFromUrls(keys)
+	if catalog == nil {
+		return
+	}
+
+	for i := 0; i < len(catalog); i++ {
```
sebrandon1, line 264: I don't think you use `i` here correct?  Is this additional for-loop needed?

pkg/tnf/identifier/identifier_test.go
```diff
@@ -83,3 +83,41 @@ func TestIdentifier_UnmarshalJSON(t *testing.T) {
 		}
 	}
 }
+
+func TestXformToGinkgoItIdentifier(t *testing.T) {
```
sebrandon1, line 87: This works but you could potentially shorten/encompass all test cases in something like:

```
testCases := []struct{}{}

for _, tc := range testCases { } 
```

cmd/tnf/generate/catalog/catalog.go
sebrandon1: I don't think you need this at all if you are already range'ing through the `catalog`.

jgato: yeah, you are right. This is because I am adding to each testName an array of identifiers. But, looking to the package, the test cases building blocks, each test has only one identifier.

pkg/tnf/identifier/identifier_test.go
jgato: Ok, I have combined both wrong and ok test, including in the tests structure the expected result of each one.

cmd/tnf/generate/catalog/catalog.go
jun-rh: As discussed with @jgato, we will keep this change for better readability.

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -129,20 +215,33 @@ func outputTestCases() {
...
-		fmt.Fprintf(os.Stdout, "Suggested Remediation|%s\n", strings.ReplaceAll(identifiers.Catalog[k].Remediation, "\n", " "))
-		fmt.Fprintf(os.Stdout, "Best Practice Reference|%s\n", strings.ReplaceAll(identifiers.Catalog[k].BestPracticeReference, "\n", " "))
+		fmt.Fprintf(os.Stdout, "### %s\n", suite)
+		fmt.Println()
+		for _, k := range catalog[suite] {
+			fmt.Fprintf(os.Stdout, "#### %s\n", k.testName)
+			fmt.Println()
+			fmt.Println("Property|Description")
+			fmt.Println("---|---")
+			fmt.Fprintf(os.Stdout, "Test Name|%s\n", k.testName)
```
jun-rh, line 235: `Test Case Name`
Sometimes I get confused which catalog I am looking at :-)

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -129,20 +215,33 @@ func outputTestCases() {
...
+			fmt.Fprintf(os.Stdout, "Url|%s\n", k.identifier.Url)
```
jun-rh, line 236: As discussed, something like global id or unique id instead of url

cmd/tnf/generate/catalog/catalog.go
jgato: should it also be Test Case Name in the other catalogue (Test Case Building Block section)?

jgato: done it

cmd/tnf/generate/catalog/catalog.go
jgato: done it

jgato:
With the last commit I have simplified very much the outputTestCaseBuildingBlocks. Sorry, I did it much more complex than need it. Because I followed the same approach than outputTestCases work, which is a little bit more complex because of the two dimensions SuiteName/TestName. 
So now outputTestCaseBuildingBlocks is 99% the same than before, but just printing the section with just a testName that you can get from the previous URL and the new added function XformToGinkgoItIdentifier.

sebrandon1 (approved):

pkg/tnf/identifier/identifier.go
```diff
@@ -85,3 +88,31 @@ func (i *Identifier) UnmarshalJSON(b []byte) error {
 
 	return i.unmarshalSemanticVersion(objMap)
 }
+
+// XformToGinkgoItIdentifier transform the Identifier into a test Id that can be used to skip
+// specific tests
+func XformToGinkgoItIdentifier(identifier Identifier) string {
```
jun-rh, line 94: As we discussed, test building blocks have nothing to do with ginkgo or It. Just call it something like GetShortName()

jgato: Renamed, it had no sense the previous name. Neither had sense to have a function pointing to names to skip tests.

jun-rh (approved):

PR #565, opened 2022-01-07
Limit the number of local shell sessions to 1

jun-rh:


test-network-function/networking/suite.go
```diff
@@ -108,6 +108,7 @@ var _ = ginkgo.Describe(common.NetworkingTestKey, func() {
 		})
 
 		ginkgo.ReportAfterEach(results.RecordResult)
+		ginkgo.AfterEach(env.CloseLocalShellContext)
```
hamadise, line 111: why not call this once, if common/suite.go ?

jun-rh: In case some command hangs, we don't want to impact all other test cases by using the same shell.

javierpena (commented):
Looks ok to me so far. I'm trying to find out why some operator tests were skipped, it looks like some corner case in the cnf-certification-test-partner repo.

javierpena:
I've finally managed to work around the operator skipped tests with https://github.com/test-network-function/cnf-certification-test-partner/pull/76 . There's also an issue with minikube (https://github.com/kubernetes/minikube/issues/11608), but it does not seem to happen when using the docker driver.

javierpena (approved):

sebrandon1 (approved):

PR #566, opened 2022-01-10
Check if port that listening is declared

shimritproj:
This test checks if all ports listening is declared.


test-network-function/networking/suite.go
```diff
@@ -323,3 +335,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+			}
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+
+			commandlisten := []string{utils.AddNsenterPrefix(containerPID), commandportlisten}
+
+			listeningPort = getListeningPortList(commandlisten, nodeOc.Context, listeningPort)
+			// compare between declaredPort,listeningPort and return the common.
+			res := checkIfListenIsDeclared(listeningPort, declaredPort)
+			if !res {
+				ginkgo.Fail("TC failed.")
```
jun-rh, line 429: Need more info here, e.g. which port is listening but not declared.

test-network-function/networking/suite.go
```diff
@@ -323,3 +335,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+		if port == "" || name == "" || protocolName == "" {
+			continue
+		}
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		name = strings.TrimSpace(strings.ReplaceAll(name, "\"", ""))
+		protocolName = strings.TrimSpace(strings.ReplaceAll(protocolName, "\"", ""))
+		declaredPort[p] = name + " " + protocolName
+		port, name, protocolName = "", "", ""
+	}
+	return declaredPort
```
sebrandon1, line 368: So `declaredPort` and the below `listeningPort` are maps which are pass-by-reference.  Something bothers me about returning a map that is already declared outside the function.  Could you declare a local-to-the-function variable to return instead of the maps brought in by parameter?

test-network-function/networking/suite.go
```diff
@@ -323,3 +335,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+		if !ok {
+			return false
+		}
+	}
+	return true
+}
+
+func testListenAndDeclared(env *config.TestEnvironment) {
+	declaredPort := make(map[int]string)
+	listeningPort := make(map[int]string)
```
jun-rh, line 400: make them plural

test-network-function/networking/suite.go
```diff
@@ -323,3 +335,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+		}
+		if port == "" || name == "" || protocolName == "" {
+			continue
+		}
+		p, _ := strconv.Atoi(strings.TrimSpace(port))
+		name = strings.TrimSpace(strings.ReplaceAll(name, "\"", ""))
+		protocolName = strings.TrimSpace(strings.ReplaceAll(protocolName, "\"", ""))
+		declaredPort[p] = name + " " + protocolName
+		port, name, protocolName = "", "", ""
+	}
```
jun-rh, line 367: All this parsing should be in a different function with unit test.

sebrandon1: Agreed.

test-network-function/networking/suite.go
shimritproj: I agree, fixed.

test-network-function/networking/suite.go
```diff
@@ -323,3 +335,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+		listeningPort[p] = fields[indexprotocolname]
+	}
+	return listeningPort
+}
+
+func checkIfListenIsDeclared(listeningPort, declaredPort map[int]string) bool {
+	for k := range listeningPort {
+		_, ok := declaredPort[k]
+		if !ok {
+			return false
```
jun-rh, line 392: where do you compare protocol?

test-network-function/networking/suite.go
```diff
@@ -323,3 +335,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+	res, _ := utils.ExecuteCommand(listeningPortCommand, ocCommandTimeOut, nodeOc)
+	lines := strings.Split(res, "\n")
+	for _, line := range lines {
+		fields := strings.Fields(line)
+		if indexprotocolname > len(fields) || indexport > len(fields) {
+			return listeningPort
+		}
+		s := strings.Split(fields[indexport], ":")
+		p, _ := strconv.Atoi(s[1])
+		listeningPort[p] = fields[indexprotocolname]
```
jun-rh, line 383: consider using a composite key [int, string]. otherwise, not sure how you can handle the same port number listening on both tcp and udp ...

test-network-function/networking/suite.go
shimritproj: OK, I fixed.

test-network-function/networking/suite.go
shimritproj: right, I fixed.

test-network-function/networking/suite.go
shimritproj: Agree, I fixed.

test-network-function/networking/suite.go
shimritproj: I agree that it's better to use a local variable with a different name but - if I do, then I need to merge the 2 maps and that's another code. It's intentionally the thing I send the "declaredPorts" and constantly updating the map.(Same for "listeningPorts")

shimritproj: I fixed it.

sebrandon1:
Looks like some previous commits got pulled in to your changes.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+			}
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+
+			commandlisten := []string{utils.AddNsenterPrefix(containerPID), commandportlisten}
+
+			listeningPortList(commandlisten, nodeOc.Context, listeningPorts)
+			// compare between declaredPort,listeningPort and return the common.
+			res := checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			if !res {
+				ginkgo.Fail("TC failed : port is listening but not declared.")
```
jun-rh, line 442: I meant actually reporting which listening ports were not declared

jun-rh: You can use the ginkgo log writer function to dump the bad ports while finding it. It's probably easier than collecting everything and dumping them at the end.

jun-rh: Also, we don't call fail on the first failure. Need to finish testing other pods

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,99 @@ func testNodePort(env *config.TestEnvironment) {
...
+
+func checkIfListenIsDeclared(listeningPorts, declaredPorts map[key]string) bool {
+	if len(listeningPorts) == 0 || len(declaredPorts) == 0 {
+		return false
+	}
+	for k := range listeningPorts {
+		_, ok := declaredPorts[k]
+		if !ok {
+			return false
+		}
```
jun-rh, line 403: Same here, finish checking all ports instead of returning false on first one

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,99 @@ func testNodePort(env *config.TestEnvironment) {
 		}
 	})
 }
+
+func parseVariables(res string, declaredPorts map[key]string) error {
```
sebrandon1, line 352: We should have a unit test surrounding this func.

test-network-function/networking/suite.go
shimritproj: Fixed.

test-network-function/networking/suite.go
shimritproj: Fixed, I was replace the "gingko.Fail" to "log.info", if you prefer something else tell me. I was search gingko log but I didn't find.

test-network-function/networking/suite.go
shimritproj: I added - log.Info("The parser is Fail.") and we will return the error, would you prefer that if it failed then the whole test failed and not continue?

test-network-function/networking/suite.go
```diff
@@ -396,13 +400,15 @@ func checkIfListenIsDeclared(listeningPorts, declaredPorts map[key]string) bool
 	if len(listeningPorts) == 0 || len(declaredPorts) == 0 {
 		return false
 	}
+	flag := true
 	for k := range listeningPorts {
 		_, ok := declaredPorts[k]
 		if !ok {
-			return false
+			log.Info(fmt.Sprintf("The port %d on protocol %s is not declared.", k.port, k.protocol))
```
jun-rh, line 407: This goes to log only. We need it in the claim. Use tnf.ClaimFilePrintf() instead.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,105 @@ func testNodePort(env *config.TestEnvironment) {
...
+	return err
+}
+
+func listeningPortList(commandlisten []string, nodeOc *interactive.Context, listeningPorts map[key]string) {
+	var k key
+	listeningPortCommand := strings.Join(commandlisten, " ")
+	fmt.Println(listeningPortCommand)
+	res, _ := utils.ExecuteCommand(listeningPortCommand, ocCommandTimeOut, nodeOc)
+	if res == "" {
+		log.Info("The command of listening ports is fail.")
```
jun-rh, line 382: Same here, err should be kept from invoking the command and printed out if not nil.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,105 @@ func testNodePort(env *config.TestEnvironment) {
...
+	declaredPorts := make(map[key]string)
+	listeningPorts := make(map[key]string)
+	var x *configsections.Container
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				err := declaredPortList(i, podUnderTest.Name, podUnderTest.Namespace, declaredPorts)
+				if err != nil {
+					return
```
jun-rh, line 424: return ends the test ...
log and continue instead

test-network-function/networking/suite.go
jun-rh: Actually, this function should just return the err if anything goes wrong. The caller can log the error and proceed accordingly.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,105 @@ func testNodePort(env *config.TestEnvironment) {
...
+			test, err := tnf.NewTest(context.GetExpecter(), nodeName, []reel.Handler{nodeName}, context.GetErrorChannel())
+			gomega.Expect(err).To(gomega.BeNil())
+			test.RunAndValidate()
+			nodeOc := env.NodesUnderTest[nodeName.GetNodeName()].DebugContainer.GetOc()
+			for _, cut := range env.ContainersUnderTest {
+				if cut.ContainerIdentifier.PodName == podUnderTest.Name && cut.ContainerIdentifier.Namespace == podUnderTest.Namespace {
+					x = cut
+					break
+				}
+			}
```
jun-rh, line 439: This loop is no longer needed. David recently added containerList to podUnderTest. Just use the first one: `podUnderTest.ContainerList[0]`

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,105 @@ func testNodePort(env *config.TestEnvironment) {
...
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+
+			commandlisten := []string{utils.AddNsenterPrefix(containerPID), commandportlisten}
+
+			listeningPortList(commandlisten, nodeOc.Context, listeningPorts)
+			// compare between declaredPort,listeningPort and return the common.
+			res := checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			if !res {
+				log.Info("TC failed : port is listening but not declared.")
```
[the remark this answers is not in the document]
shimritproj: fixed.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,105 @@ func testNodePort(env *config.TestEnvironment) {
...
+		declaredPorts[k] = p[element].Name
+	}
+	return nil
+}
+func declaredPortList(container int, podName, podNamespace string, declaredPorts map[key]string) error {
+	ocCommandToExecute := fmt.Sprintf(commandportdeclared, podName, podNamespace, container)
+	res, _ := utils.ExecuteCommand(ocCommandToExecute, ocCommandTimeOut, interactive.GetContext(false))
+	err := parseVariables(res, declaredPorts)
+	if err != nil {
+		log.Info("The parser is Fail.")
```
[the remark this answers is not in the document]
shimritproj: I changed to- log.Error("ERROR: The parser is Fail.").

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,111 @@ func testNodePort(env *config.TestEnvironment) {
...
+	declaredPorts := make(map[key]string)
+	listeningPorts := make(map[key]string)
+	undeclaredPorts, res := make(map[key]string), false
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				err := declaredPortList(i, podUnderTest.Name, podUnderTest.Namespace, declaredPorts)
+				if err != nil {
+					log.Error(err)
```
jun-rh, line 427: Log message should contain pod info so we know which one failed

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,111 @@ func testNodePort(env *config.TestEnvironment) {
...
+			test.RunAndValidate()
+			nodeOc := env.NodesUnderTest[nodeName.GetNodeName()].DebugContainer.GetOc()
+			x := podUnderTest.ContainerList[0]
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+
+			commandlisten := []string{utils.AddNsenterPrefix(containerPID), commandportlisten}
+
+			err = listeningPortList(commandlisten, nodeOc.Context, listeningPorts)
+			if err != nil {
+				log.Error(err)
```
jun-rh, line 445: same here

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,111 @@ func testNodePort(env *config.TestEnvironment) {
...
+	if err != nil {
+		log.Error(err)
+	}
+	return err
+}
+
+func listeningPortList(commandlisten []string, nodeOc *interactive.Context, listeningPorts map[key]string) error {
+	var k key
+	listeningPortCommand := strings.Join(commandlisten, " ")
+	fmt.Println(listeningPortCommand)
```
jun-rh, line 379: debug log?

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,111 @@ func testNodePort(env *config.TestEnvironment) {
...
+	res, err := utils.ExecuteCommand(listeningPortCommand, ocCommandTimeOut, nodeOc)
+	if res == "" {
```
jun-rh, line 381:
```suggestion
	if err != nil {
```

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,111 @@ func testNodePort(env *config.TestEnvironment) {
...
+		log.Error(err)
```
jun-rh, line 382: No need to log if err is returned

test-network-function/networking/suite.go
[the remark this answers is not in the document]
jun-rh: This is not fixed. Use tnf.ClaimFilePrintf() and include Pod info.

test-network-function/networking/suite.go
shimritproj: I was delete this line : fmt.Println(listeningPortCommand)

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,109 @@ func testNodePort(env *config.TestEnvironment) {
...
+	declaredPorts := make(map[key]string)
+	listeningPorts := make(map[key]string)
+	undeclaredPorts, res := make(map[key]string), false
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				err := declaredPortList(i, podUnderTest.Name, podUnderTest.Namespace, declaredPorts)
+				if err != nil {
+					log.Errorf("Failed: There are no declared ports for pod name %s in pod namespace %s", podUnderTest.Name, podUnderTest.Namespace)
```
jun-rh, line 425:
```suggestion
					log.Errorf("Failed to get declared ports for pod name %s in pod namespace %s due to %v", podUnderTest.Name, podUnderTest.Namespace, err)
```

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,109 @@ func testNodePort(env *config.TestEnvironment) {
...
+	}
+	return nil
+}
+
+func checkIfListenIsDeclared(listeningPorts, declaredPorts map[key]string) (flag bool, res map[key]string) {
+	res = make(map[key]string)
+	if len(listeningPorts) == 0 || len(declaredPorts) == 0 {
+		return false, res
+	}
+	flag = true
```
jun-rh, line 403: no longer needed now that you are returning the port list. success if list is empty.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,109 @@ func testNodePort(env *config.TestEnvironment) {
...
+			}
+			// compare between declaredPort,listeningPort and return the common.
+			res, undeclaredPorts = checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			if !res {
+				tnf.ClaimFilePrintf("TC failed : port is listening but not declared in pod name %s and pod namespace is %s", podUnderTest.Name, podUnderTest.Namespace)
+			}
+			for k := range undeclaredPorts {
+				tnf.ClaimFilePrintf("The port %d on protocol %s in pod name %s and pod namespace is %s not declared.", k.port, k.protocol, podUnderTest.Name, podUnderTest.Namespace)
+			}
+		}
```
jun-rh, line 454: at the end of the It block (outside the loop), we still need to call ginkgo fail or error depending on what we find in the loop. if we got undeclaredPorts from any pod, or we failed to retrieve declared ports or listening port on any pod, call ginkgo.fail().

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,109 @@ func testNodePort(env *config.TestEnvironment) {
...
+				continue
+			}
+			// compare between declaredPort,listeningPort and return the common.
+			res, undeclaredPorts = checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			if !res {
+				tnf.ClaimFilePrintf("TC failed : port is listening but not declared in pod name %s and pod namespace is %s", podUnderTest.Name, podUnderTest.Namespace)
+			}
+			for k := range undeclaredPorts {
+				tnf.ClaimFilePrintf("The port %d on protocol %s in pod name %s and pod namespace is %s not declared.", k.port, k.protocol, podUnderTest.Name, podUnderTest.Namespace)
+			}
```
jun-rh, line 448:
```suggestion
                            failedPods = append(failedPods, podUnderTest)
			}
```
Check failedPods outside of the loop

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,109 @@ func testNodePort(env *config.TestEnvironment) {
...
+			err = listeningPortList(commandlisten, nodeOc.Context, listeningPorts)
+			if err != nil {
+				log.Errorf("Failed: There are no listening ports for pod name %s in pod namespace %s", podUnderTest.Name, podUnderTest.Namespace)
+				continue
+			}
+			// compare between declaredPort,listeningPort and return the common.
+			res, undeclaredPorts = checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			if !res {
+				tnf.ClaimFilePrintf("TC failed : port is listening but not declared in pod name %s and pod namespace is %s", podUnderTest.Name, podUnderTest.Namespace)
+			}
```
jun-rh:
```suggestion
                            failedPods = append(failedPods, podUnderTest)
			}
```
Check failedPods outside of the loop

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,109 @@ func testNodePort(env *config.TestEnvironment) {
...
+	listeningPorts := make(map[key]string)
+	undeclaredPorts, res := make(map[key]string), false
+	testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestServicesDoNotUseNodeportsIdentifier)
+	ginkgo.It(testID, func() {
+		for _, podUnderTest := range env.PodsUnderTest {
+			for i := 0; i < podUnderTest.ContainerCount; i++ {
+				err := declaredPortList(i, podUnderTest.Name, podUnderTest.Namespace, declaredPorts)
+				if err != nil {
+					log.Errorf("Failed: There are no declared ports for pod name %s in pod namespace %s", podUnderTest.Name, podUnderTest.Namespace)
+					continue
```
jun-rh, line 424: when we fail to get declared ports from one of the containers, there is no point to continue with this pod.

- Add this pod to a skipped list so we can check at the end before calling ginkgo.Fail or gomega.Assert(len(skippedPods), 0). 
- This "continue" statement needs to be on the outer for loop (line 421). See [this reference](https://relistan.com/continue-statement-with-labels-in-go)

shimritproj: you're right, I fixed it.

test-network-function/networking/suite.go
```diff
@@ -395,35 +395,37 @@ func listeningPortList(commandlisten []string, nodeOc *interactive.Context, list
...
 	ginkgo.It(testID, func() {
+	OUTER:
 		for _, podUnderTest := range env.PodsUnderTest {
 			for i := 0; i < podUnderTest.ContainerCount; i++ {
 				err := declaredPortList(i, podUnderTest.Name, podUnderTest.Namespace, declaredPorts)
 				if err != nil {
-					log.Errorf("Failed: There are no declared ports for pod name %s in pod namespace %s", podUnderTest.Name, podUnderTest.Namespace)
-					continue
+					log.Errorf("Failed to get declared ports for container #%d pod name %s in pod namespace %s due to %v", i, podUnderTest.Name, podUnderTest.Namespace, err)
+					skippedPods[podUnderTest.Name] = podUnderTest.Namespace
```
jun-rh, line 427: Namespace?

test-network-function/networking/suite.go
```diff
@@ -440,17 +442,23 @@ func testListenAndDeclared(env *config.TestEnvironment) {
...
+			// compare between declaredPort,listeningPort
+			undeclaredPorts = checkIfListenIsDeclared(listeningPorts, declaredPorts)
 			for k := range undeclaredPorts {
 				tnf.ClaimFilePrintf("The port %d on protocol %s in pod name %s and pod namespace is %s not declared.", k.port, k.protocol, podUnderTest.Name, podUnderTest.Namespace)
 			}
+			if len(undeclaredPorts) != 0 {
+				failedPods[podUnderTest.Name] = podUnderTest.Namespace
+			}
+		}
+		for k := range failedPods {
```
jun-rh, line 457: Fail only works once. No need to use a for loop unless you want to build a Fail message with the content from the pod list.

test-network-function/networking/suite.go
shimritproj: I fixed it.

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,117 @@ func testNodePort(env *config.TestEnvironment) {
...
+			if len(undeclaredPorts) != 0 {
+				failedPods = append(failedPods, podUnderTest.Name)
+			}
+		}
+		if n := len(failedPods); n > 0 {
+			ginkgo.Fail("Failed to get declared ports for pod %s", n)
+		}
+		if n := len(skippedPods); n > 0 {
+			ginkgo.Fail("Skipped pod to get declared ports for pod %s", n)
+		}
```
jun-rh, line 462:
```suggestion
		if nf := len(failedPods); ns := len(skippedPods); nf > 0 || ns >0 {
			ginkgo.Fail("Found %d pods with listening ports not declared and Skipped %d pods due to unexpected error", nf, ns)
		}
```

test-network-function/networking/suite.go
```diff
@@ -324,3 +348,115 @@ func testNodePort(env *config.TestEnvironment) {
...
+	}
+	return res
+}
+
+func testListenAndDeclared(env *config.TestEnvironment) {
+	declaredPorts := make(map[key]string)
+	listeningPorts := make(map[key]string)
+	undeclaredPorts := make(map[key]string)
+	var skippedPods []string
```
jun-rh, line 417: we don't use the content of these lists now, just the length. so it's kind of ok. but it really should just be []Pod instead of []string.

jun-rh (approved):

sebrandon1 (approved):

test-network-function/networking/suite.go
shimritproj: you're right! I fixed.

test-network-function/networking/suite.go
shimritproj: After the meeting, I close this , I will create new PR that include unit test for all function in file suite.go in networking.

PR #571, opened 2022-01-11
Use labels in test cases

javierpena:
This commit adds ginkgo labels [1] to all test cases, so we can use
filtering to select which test case to run.

In addition to this, the run-cnf-suites.sh script is adapted to allow
filtering by label.

[1] - https://onsi.github.io/ginkgo/#spec-labels


jun-rh (approved):

jun-rh:
Now we can actually put the descriptive text back as 'It'. We had some and got lost when @hamadise had to make them usable by "-s"

run-cnf-suites.sh
```diff
@@ -88,11 +95,16 @@ fi
 
 echo "Running with focus '$FOCUS'"
 echo "Running with skip  '$SKIP'"
+echo "Running with label filter '$LABEL'"
 echo "Report will be output to '$OUTPUT_LOC'"
 echo "ginkgo arguments '${GINKGO_ARGS}'"
 SKIP_STRING=""
+LABLE_STRING=""
```
edcdavid, line 102: typo

test-network-function/networking/suite.go
```diff
@@ -202,7 +202,7 @@ func runNetworkingTests(netsUnderTest map[string]netTestContext, count int) map[
 func testDefaultNetworkConnectivity(env *config.TestEnvironment, count int) {
 	ginkgo.When("Testing Default network connectivity", func() {
 		testID := identifiers.XformToGinkgoItIdentifier(identifiers.TestICMPv4ConnectivityIdentifier)
-		ginkgo.It(testID, func() {
+		ginkgo.It(testID, ginkgo.Label(testID), func() {
```
edcdavid, line 205: This test should not have the same label as the multus test. The test is also defined with the context hierarchy e.g. with ginkgo.When.
there is a function to do this automatically (used for claim.json) but it would make the label too long -> ContainerHierarchyTexts

See:
networking Both Pods are on the Default network Testing Default network connectivity
  networking-icmpv4-connectivity [**networking-icmpv4-connectivity**]
  /home/deliedit/redhat/github/david/upstream/test-network-function/networking/suite.go:205
networking Both Pods are connected via a Multus Overlay Network Testing Multus network connectivity
  networking-icmpv4-connectivity [**networking-icmpv4-connectivity**]

hamadise:
> Now we can actually put the descriptive text back as 'It'. We had some and got lost when @hamadise had to make them usable by "-s"

Right, The descriptive test did mostly move to a `By` block. Partners are using it now to define which test did pass or not. I would not mind bringing back description to It block

run-cnf-suites.sh
javierpena: Ouch! Thanks for the catch.

test-network-function/networking/suite.go
javierpena: Actually, shouldn't they use a different test id? If that is the only case, it might be easier to change the multus test id to be something like networking-icmpv4-connectivity-multus

greyerof (approved):
I agree with Javier here. The multus networking TC should have its own testID, but that can be done in a new PR.

jun-rh:
Issue #576 created. @javierpena or @edcdavid, feel free to grab it.

javierpena:
> Issue #576 created. @javierpena or @edcdavid, feel free to grab it.

I'll take care of it.

PR #575, opened 2022-01-13
Fix false positives caused by the grep expression in clusterrolebinding.go

jun-rh:
Original:
image
Fix:
image


javierpena (approved):

sebrandon1 (approved):
This looks similar to the latest Mavenir results.  Approved.

PR #577, opened 2022-01-13
Update multus networking TC id

### LINKED ISSUE #576: Update multus networking TC id
See discussion in #571

javierpena:
Previously, the networking ICMPv4 connectivity tests for the default and
Multus networks shared the same id. This commit splits that, so each
test case has a single id.

Fixes #576

### DIFF
--- a/test-network-function/identifiers/identifiers.go
+++ b/test-network-function/identifiers/identifiers.go
@@ -97,8 +97,8 @@ var (
 		Url:     formTestURL(common.NetworkingTestKey, "icmpv4-connectivity"),
 		Version: versionOne,
 	}
-	// TestICMPv4ConnectivityIdentifierMultus tests icmpv4 connectivity on multus networks.
-	TestICMPv4ConnectivityIdentifierMultus = claim.Identifier{
+	// TestICMPv4ConnectivityMultusIdentifier tests icmpv4 connectivity on multus networks.
+	TestICMPv4ConnectivityMultusIdentifier = claim.Identifier{
 		Url:     formTestURL(common.NetworkingTestKey, "icmpv4-connectivity-multus"),
 		Version: versionOne,
 	}
@@ -354,13 +354,13 @@ test case requires the Deployment of the debug daemonset.`),
 		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 6.2",
 	},
 
-	TestICMPv4ConnectivityIdentifierMultus: {
-		Identifier: TestICMPv4ConnectivityIdentifierMultus,
+	TestICMPv4ConnectivityMultusIdentifier: {
+		Identifier: TestICMPv4ConnectivityMultusIdentifier,
 		Type:       normativeResult,
 		Remediation: `Ensure that the CNF is able to communicate via the Multus network(s). In some rare cases,
 CNFs may require routing table changes in order to communicate over the Multus network(s). To exclude a particular pod
 from ICMPv4 connectivity tests, add the test-network-function.com/skip_connectivity_tests label to it. The label value is not important, only its presence.`,
-		Description: formDescription(TestICMPv4ConnectivityIdentifierMultus,
+		Description: formDescription(TestICMPv4ConnectivityMultusIdentifier,
 			`checks that each CNF Container is able to communicate via ICMPv4 on the Multus network(s).  This
 test case requires the Deployment of the debug daemonset.`),
 		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 6.2",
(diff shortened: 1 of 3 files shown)


jun-rh (approved):

test-network-function/identifiers/identifiers.go
```diff
@@ -333,6 +338,20 @@ test case requires the Deployment of the debug daemonset.
 		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 6.2",
 	},
 
+	TestICMPv4ConnectivityIdentifierMultus: {
+		Identifier: TestICMPv4ConnectivityIdentifierMultus,
+		Type:       normativeResult,
+		Remediation: `Ensure that the CNF is able to communicate via the Multus network(s).  In some rare cases,
+CNFs may require routing table changes in order to communicate over the Multus network(s).  In other cases, if the
+Container base image does not provide the "ip" or "ping" binaries, this test may not be applicable.  For instructions on
```
greyerof, line 346: Since we started using nsenter, the "ping" binary is no longer needed to be inside the container image. If I'm not wrong, nsenter uses the binaries from the host where the tnf is running. This is applicable for both networking TCs (default+multus).

javierpena: I think you're right, I'll fix that.

test-network-function/identifiers/identifiers.go
```diff
@@ -333,6 +337,19 @@ test case requires the Deployment of the debug daemonset.
 		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 6.2",
 	},
 
+	TestICMPv4ConnectivityIdentifierMultus: {
+		Identifier: TestICMPv4ConnectivityIdentifierMultus,
+		Type:       normativeResult,
+		Remediation: `Ensure that the CNF is able to communicate via the Multus network(s). In some rare cases,
+CNFs may require routing table changes in order to communicate over the Multus network(s). For instructions on
+how to exclude a particular container from ICMPv4 connectivity tests, consult:
+[README.md](https://github.com/test-network-function/test-network-function#issue-161-some-containers-under-test-do-not-contain-ping-or-ip-binary-utilities).`,
```
jun-rh, line 346: Bad link. We've removed this from README.

jun-rh: We still have one sentence there mentioning the exclude label. Not sure we need to link that or maybe just mention it here too.

javierpena: I'll just add the text, I think it makes sense. Thanks!

test-network-function/identifiers/identifiers.go
```diff
@@ -321,15 +326,24 @@ they are the same.`),
...
-`),
+test case requires the Deployment of the debug daemonset.`),
+		BestPracticeReference: bestPracticeDocV1dot2URL + " Section 6.2",
+	},
+
+	TestICMPv4ConnectivityIdentifierMultus: {
+		Identifier: TestICMPv4ConnectivityIdentifierMultus,
+		Type:       normativeResult,
+		Remediation: `Ensure that the CNF is able to communicate via the Multus network(s). In some rare cases,
+CNFs may require routing table changes in order to communicate over the Multus network(s). To exclude a particular container
```
jun-rh, line 342: Label is on the pod. We used to support skipping specific container with static config. That got dropped at some point.

test-network-function/identifiers/identifiers.go
greyerof: So I was wrong. It uses the debug pod's context to issue the nsenter command. Thanks @jun-rh .

test-network-function/identifiers/identifiers.go
javierpena: Fixed, thanks!

test-network-function/identifiers/identifiers.go
```diff
@@ -97,6 +97,11 @@ var (
 		Url:     formTestURL(common.NetworkingTestKey, "icmpv4-connectivity"),
 		Version: versionOne,
 	}
+	// TestICMPv4ConnectivityIdentifierMultus tests icmpv4 connectivity on multus networks.
+	TestICMPv4ConnectivityIdentifierMultus = claim.Identifier{
```
greyerof, line 101: I would put the "Identifier" word as suffix always.

greyerof (approved):

test-network-function/identifiers/identifiers.go
javierpena: Makes sense, fixed. Thanks!

PR #579, opened 2022-01-17
Collect and parse image source for contains under test

jun-rh:


pkg/config/autodiscover/autodiscover.go
```diff
@@ -141,6 +142,33 @@ func buildContainers(pr *PodResource) []configsections.Container {
 	return containers
 }
 
+//nolint:gomnd
+func buildContainerImageSource(url string) *configsections.ContainerImageSource {
```
sebrandon1, line 146: Any reason specifically to return a pointer here instead of the object itself?  Just curious.

sebrandon1 (approved):
One question, other than that looks good to me.

pkg/config/autodiscover/autodiscover.go
jun-rh: pkg/config/config.go:247:2: rangeValCopy: each iteration copies 184 bytes (consider pointers or indexing) (gocritic)
        for _, cid := range env.Config.Partner.ContainersDebugList {
        ^
pkg/config/config.go:345:2: rangeValCopy: each iteration copies 184 bytes (consider pointers or indexing) (gocritic)
        for _, debugPod := range env.Config.Partner.ContainersDebugList {
        ^
pkg/config/config.go:357:2: rangeValCopy: each iteration copies 184 bytes (consider pointers or indexing) (gocritic)
        for _, c := range containerDefinitions {

PR #581, opened 2022-01-18
Include discovered container images in the container certification test

jun-rh:


test-network-function/certification/suite.go
```diff
@@ -87,25 +87,32 @@ func waitForCertificationRequestToSuccess(certificationRequestFunc func() bool,
...
+		}
+		if env.Config.CheckDiscoveredContainerCertificationStatus {
+			const redhatContainerCatalogRegistry = "https://catalog.redhat.com/software/containers/"
+			for _, cut := range env.ContainersUnderTest {
+				if cut.ImageSource.Registry != redhatContainerCatalogRegistry {
+					containersToQuery[configsections.CertifiedContainerRequestInfo{Repository: cut.ImageSource.Repository, Name: cut.ImageSource.Name}] = true
+				}
+			}
+		}
+		if len(containersToQuery) == 0 && !env.Config.CheckDiscoveredContainerCertificationStatus {
```
javierpena, line 107: If I understood the logic here correctly, it should be || (or), right?

jun-rh: no, we call skip only if there is no query to make and we didn't check (based on prefix/registry) discovered ones. i will add a comment to the prefix check. that is being considered success result too.

README.md
```diff
@@ -85,9 +85,12 @@ the subscription for this CSV. If unset, the CSV name will be used.
 
 ### certifiedcontainerinfo and certifiedoperatorinfo
 
-The `certifiedcontainerinfo` and `certifiedoperatorinfo` sections contain information about CNFs and Operators that are
+The `certifiedcontainerinfo` and `certifiedoperatorinfo` sections contain information about Containers and Operators that are
 to be checked for certification status on Red Hat catalogs.
 
+### checkDiscoveredContainerCertificationStatus
+This boolean flag can be turned on when you intent to have the test suite check the certification status of the container images used by the autodiscoverd test target pods
```
edcdavid, line 92: add: " in addition to the configured image list."

test-network-function/certification/suite.go
```diff
@@ -87,25 +87,32 @@ func waitForCertificationRequestToSuccess(certificationRequestFunc func() bool,
...
 		env := configpkg.GetTestEnvironment()
-		containersToQuery := env.Config.CertifiedContainerInfo
-
-		if len(containersToQuery) == 0 {
+		containersToQuery := make(map[configsections.CertifiedContainerRequestInfo]bool)
+		for _, c := range env.Config.CertifiedContainerInfo {
+			containersToQuery[c] = true
+		}
+		if env.Config.CheckDiscoveredContainerCertificationStatus {
+			const redhatContainerCatalogRegistry = "https://catalog.redhat.com/software/containers/"
```
edcdavid, line 100: What is the purpose of redhatContainerCatalogRegistry? the url does not look like a image registry.

jun-rh: didn't you say everything under that was certified? :smile:

edcdavid: This is the catalog for all certified containers, correct (https://catalog.redhat.com/software/containers/explore). But I believe that here we need the registry, similar to : https://catalog.redhat.com/software/containers/cp/ibm-eventstreams-schema-registry/5f583d17ac3db90370a1c55a?container-tabs=gti

jun-rh: Not sure what you mean. We used this one for a while and the url has the prefix:

https://catalog.redhat.com/software/containers/rhel8/support-tools/5ba3eaf9bed8bd6ee819b78b

jun-rh: hmm, now i see that this url doesn't use @ for digest.

jun-rh: ok, i see that the real url should be like this:

registry.access.redhat.com/rhel8/support-tools@sha256:e1ae34233f8aadbe5f5b026eadfbb6c24d30b47d2e8cc612731d074bc2ad7d00

edcdavid: yes, correct

jun-rh: My mistake. It was actually "registry.redhat.io/rhel8/support-tools:latest".

jun-rh: Maybe i will just get rid of this check? You guys can optimize it later once you confirm the certification status of the images in that registry in general.

edcdavid: yes I think we can remove it. I did a quick check on the container entries in the catalog, we can at least check based on platform type (x86, arm, ..) and docker version to improve the error message

jun-rh: Done.

test-network-function/certification/suite.go
jun-rh: This has been removed as we are not sure about the registry and we may want to do the query anyways for better verification that considers other things like version platform etc.

edcdavid (approved):

test-network-function/certification/suite.go
```diff
@@ -87,25 +87,29 @@ func waitForCertificationRequestToSuccess(certificationRequestFunc func() bool,
...
 		env := configpkg.GetTestEnvironment()
-		containersToQuery := env.Config.CertifiedContainerInfo
-
+		containersToQuery := make(map[configsections.CertifiedContainerRequestInfo]bool)
+		for _, c := range env.Config.CertifiedContainerInfo {
+			containersToQuery[c] = true
+		}
+		if env.Config.CheckDiscoveredContainerCertificationStatus {
+			for _, cut := range env.ContainersUnderTest {
+				containersToQuery[configsections.CertifiedContainerRequestInfo{Repository: cut.ImageSource.Repository, Name: cut.ImageSource.Name}] = true
```
greyerof, line 101: Aren't ImageSource.Tag/Digest needed here?

javierpena (approved):

PR #586, opened 2022-01-20
Ensure catalog suites are always ordered

### LINKED ISSUE #583: catalog generator no longer keeps the suites in order
this is probably broken by #561

javierpena:
Previously, we were not ensuring an order when getting the list of
suites in outputTestCases(). This caused issues, because the CATALOG.md
file could be changed by any PR, breaking order.

Fixes #583

### DIFF
--- a/cmd/tnf/generate/catalog/catalog.go
+++ b/cmd/tnf/generate/catalog/catalog.go
@@ -201,6 +201,9 @@ func outputTestCases() {
 	// we need the list of suite's names
 	suites := getSuitesFromIdentifiers(keys)
 
+	// Sort the list of suite names
+	sort.Strings(suites)
+
 	// Iterating the map by test and suite names
 	for _, suite := range suites {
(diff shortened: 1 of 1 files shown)


jun-rh (approved):

cmd/tnf/generate/catalog/catalog.go
```diff
@@ -201,6 +201,9 @@ func outputTestCases() {
 	// we need the list of suite's names
 	suites := getSuitesFromIdentifiers(keys)
 
+	// Sort the list of suite names
+	sort.Strings(suites)
```
edcdavid, line 205: Shoudn't we also checkin the sorted catalog.md?

edcdavid (approved):

cmd/tnf/generate/catalog/catalog.go
jun-rh: it's sorted right now. we've been careful about messing it up since the bug was introduced :smile:

PR #591, opened 2022-01-21
Diagnostic TS removal

greyerof:
Non of the existing TCs in the diagnostic test suite is actually a TC,
since they cannot fail. They're just retrieving information from the
cluster, so the only way for them to fail comes from network or cluster
permissions related errors. On those cases, the output of any other
TS/TC would also be unreliable/unstrustable.

This PR makes the diagnostic retrieval to be done right before any other
TS, so the claim.json will always have that information, no matter which
TS was launched.

~~A new parameter "-diagnostic" will be parsed by the application to allow
the execution of the diagnostic code and dump it to the claim file.~~


sebrandon1 (approved):

.github/workflows/pre-main.yaml
```diff
@@ -188,8 +188,8 @@ jobs:
           cp test-network-function/*.yml $TNF_CONFIG_DIR
         shell: bash
 
-      - name: 'Test: Run diagnostic test suite in a TNF container'
-        run: ./run-tnf-container.sh ${{ env.TESTING_CMD_PARAMS }} -f diagnostic
+      - name: 'Test: Run without any TS, just get diagnostic information'
+        run: ./run-tnf-container.sh ${{ env.TESTING_CMD_PARAMS }}
```
javierpena, line 192: Thinking about it, we're somehow hiding this parameter under a non-documented behavior. What about making it explicit in the script? Something like having the user input one of the following options:

--focus focus1, focus2 (same as before)
--all-focus (to run all tests, similar to -f "" currently)
--diagnostic (to just run the diagnostics)

That would require some more work in the script itself, but would allow us to be more explicit in what we want to run.

greyerof: You're right, but, after studying this, it would involve more changes in all the launching scripts (which currently use slightly different syntax), which seems a bit too much for the goal of the PR. Do you mind if I just (1) add a note in the README file to document this (hidden) behavior and (2) create a repo issue about the need to update the launching scripts adding the support for the "--all-focus" and "--diagnostic" flags so it can be addressed in a different PR?

javierpena: Ok, let's do that.

javierpena (approved):

.github/workflows/pre-main.yaml
edcdavid: shouldn't we have diagnostics run all the time and just prevent suites to run if the focus string is empty?

edcdavid (approved):

PR #595, opened 2022-01-24
Add tag/digest support in container certification query

jun-rh:


internal/api/api.go
```diff
@@ -36,15 +38,33 @@ func NewHTTPClient() CertAPIClient {
...
+func (api CertAPIClient) IsContainerCertified(id configsections.ContainerImageIdentifier) (bool, error) {
+	var url string
+	const defaultTag = "latest"
+	if id.Digest == "" {
+		if id.Tag == "" {
+			id.Tag = defaultTag
+		}
+		url = fmt.Sprintf("%s/%s/%s/images?page_size=1&filter=repositories.repository==%s/%s;repositories.tags.name==%s", apiCatalogByRepositoriesBaseEndPoint, id.Repository, id.Name, id.Repository, id.Name, id.Tag)
+	} else {
+		url = fmt.Sprintf("%s/%s/%s/images?page_size=1&filter=image_id==%s", apiCatalogByRepositoriesBaseEndPoint, id.Repository, id.Name, id.Digest)
```
sebrandon1, line 59: Maybe this section should be made into its own function to properly build the URL.  Easy to unit test as well.

sebrandon1 (commented):
Just a couple of comments.

internal/api/api.go
jun-rh: Done

sebrandon1 (approved):

test-network-function/certification/suite.go
```diff
@@ -117,7 +117,7 @@ func testContainerCertificationStatus() {
 				}
 				allContainersToQueryEmpty = false
 				ginkgo.By(fmt.Sprintf("Container %s/%s should eventually be verified as certified", c.Repository, c.Name))
-				isCertified := waitForCertificationRequestToSuccess(getContainerCertificationRequestFunction(c.Repository, c.Name), apiRequestTimeout)
+				isCertified := waitForCertificationRequestToSuccess(getContainerCertificationRequestFunction(c), apiRequestTimeout)
```
edcdavid, line 120: In addition to basic certification result, maybe we could return the grade of the image and fail with recommendation to upgrade or pass with warning, if the image is too old. For instance, if the grade is >C we could fail, since older images a subject to security issues.
https://catalog.redhat.com/api/containers/v1/repositories/registry/registry.access.redhat.com/repository/rhel8/nginx-116/images?filter=freshness_grades.grade==D

edcdavid (approved):

test-network-function/certification/suite.go
jun-rh: #596 raised

PR #597, opened 2022-01-25
Add identifer tests

sebrandon1:


test-network-function/identifiers/identifiers.go
```diff
@@ -269,7 +268,7 @@ func XformToGinkgoItIdentifierExtended(identifier claim.Identifier, extra string
 // it extracts the suite name and test name from a claim.Identifier based
 // on the const url which contains a base domain
 // From a claim.Identifier.url:
-//   http://test-network-function.com/tests-case/SuitName/TestName
+//   http://test-network-function.com/testcases/SuiteName/TestName
 // It extracts SuitNAme and TestName
```
javierpena, line 272: If you're fixing the comment in line 272, it would be nice to complete the fix with line 273, too ;)

test-network-function/identifiers/identifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestGetSuiteAndTestFromIdentifier(t *testing.T) {
+	testCases := []struct {
+		testIdentifier claim.Identifier
+		expectedResult []string
+	}{
+		{
+			testIdentifier: claim.Identifier{
+				Url: "http://test-network-function.com/testcases/SuiteName/TestName/MyTest",
+			},
+			expectedResult: []string{"SuiteName", "TestName", "MyTest"},
```
greyerof, line 35: This is a weird example. What's the difference between TestName and MyTest? We should only have SuiteName and TestName here, right?. Problem here comes from GetSuiteAndTestFromIdentifier(), which I believe it should be modified to assert that after the URL prefix "http://test-network-function.com/testcases/" there's only two fields separated by "/". Also, the return value should be (tsName string, tcName string, error) to be more explicit and do exactly what the name suggests.

test-network-function/identifiers/identifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func TestXformToGinkgoItIdentifier(t *testing.T) {
+	testCases := []struct {
+		testIdentifier claim.Identifier
+		expectedResult string
+	}{
+		{
+			testIdentifier: claim.Identifier{
+				Url: "http://test-network-function.com/testcases/SuiteName/TestName/MyTest",
```
greyerof, line 58: The final "/MyTest" doesn't make sense here, either.

test-network-function/identifiers/identifier_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func TestXformToGinkgoItIdentifierExtended(t *testing.T) {
+	testCases := []struct {
+		testIdentifier claim.Identifier
+		expectedResult string
+	}{
+		{
+			testIdentifier: claim.Identifier{
+				Url: "http://test-network-function.com/testcases/SuiteName/TestName/MyTest",
```
greyerof, line 83: Same here.

test-network-function/identifiers/identifier_test.go
sebrandon1: All the function is doing is `strings.Split(result[1], "/")`.  I'm just demonstrating that it does.  It doesn't make sense but that's what it's doing.  😄

sebrandon1: The function should only return [0:2] if that's what it should be doing.  Should I add that?

jun-rh: https://github.com/test-network-function/test-network-function/issues/583#issuecomment-1018283570
@jgato Are you still planning to churn this code soon?

test-network-function/identifiers/identifier_test.go
sebrandon1: Added additional test cases and logic to remove the additional `/MyTest` from the lines.

javierpena (approved):

hamadise (approved):

PR #599, opened 2022-01-26
Add freshness grade check to container certification test

jc-rh:


internal/api/api.go
```diff
@@ -62,13 +169,15 @@ func (api CertAPIClient) IsContainerCertified(id configsections.ContainerImageId
...
 	var url string
 	const defaultTag = "latest"
+	const passingFreshnessGrades = "A,B,C"
 	if id.Digest == "" {
 		if id.Tag == "" {
 			id.Tag = defaultTag
 		}
-		url = fmt.Sprintf("%s/%s/%s/images?filter=repositories.repository==%s/%s;repositories.tags.name==%s", apiCatalogByRepositoriesBaseEndPoint, id.Repository, id.Name, id.Repository, id.Name, id.Tag)
+		url = fmt.Sprintf("%s/%s/%s/images?filter=freshness_grades.grade=in=(%s);repositories.repository==%s/%s;repositories.tags.name==%s",
+			apiCatalogByRepositoriesBaseEndPoint, id.Repository, id.Name, passingFreshnessGrades, id.Repository, id.Name, id.Tag)
```
edcdavid, line 178: In addition to the test pass/fail result, should we add the reason for failing, for instance:
- test failed because the image xxx freshness grade is D (needs to be equal or better than C, e.g. No missing Critical security errata is older than 30 days. No missing Important security errata is older than 90 days.) 
- image not certified at any point in time

jun-rh: that would require either another query or a query without the filter so we get all entries and check one by one. not sure if it's really worth it.

edcdavid: Maybe we can add it later. I was thinking of just picking the first entry found in the current filter and return that instead of len>0

edcdavid (approved):

internal/api/api.go
jun-rh: ah ok, i can try sort_by instead of the filter on the freshness grade field.

edcdavid (approved):

sebrandon1 (approved):

PR #600, opened 2022-01-26
Bump GolangCI-lint to 1.44.0

sebrandon1:
https://github.com/golangci/golangci-lint/releases/tag/v1.44.0

The v1.44.0 linter caught a line in test-network-function/platform/suite.go so I fixed that as well.

### DIFF
--- a/test-network-function/platform/suite.go
+++ b/test-network-function/platform/suite.go
@@ -696,8 +696,8 @@ func testNodeHugepagesWithMcSystemd(nodeName string, nodeNumaHugePages, mcSystem
 				}
 			}
 			if !configMatching {
-				return false, fmt.Errorf(fmt.Sprintf("MC numa=%d, hugepages (count:%d, size:%d) not matching node ones: %s",
-					mcNumaIdx, mcHugepagesCfg.hugepagesCount, mcHugepagesCfg.hugepagesSize, nodeNumaHugePages))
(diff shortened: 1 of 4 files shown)


test-network-function/platform/suite.go
```diff
@@ -696,7 +696,7 @@ func testNodeHugepagesWithMcSystemd(nodeName string, nodeNumaHugePages, mcSystem
 				}
 			}
 			if !configMatching {
-				return false, fmt.Errorf(fmt.Sprintf("MC numa=%d, hugepages (count:%d, size:%d) not matching node ones: %s",
+				return false, fmt.Errorf("%s", fmt.Sprintf("MC numa=%d, hugepages (count:%d, size:%d) not matching node ones: %s",
```
greyerof, line 699: Errorf + Sprintf seems redundant here... my fault, this was my code. But how come the linter did/does not complain? This can be left as:
```
return false, fmt.Errorf("MC numa=%d, hugepages (count:%d, size:%d) not matching node ones: %s",
    mcNumaIdx, mcHugepagesCfg.hugepagesCount, mcHugepagesCfg.hugepagesSize, nodeNumaHugePages)
```

sebrandon1: Yeah I thought so too but the linter was only complaining about the %s.  I will updated.

jun-rh (approved):

javierpena (approved):

PR #602, opened 2022-01-26
Add accesscontrol tests

sebrandon1:
* Adding unit tests.
* Removing unnecessary local variables in some funcs.
* Using `StringInSlice` to remove some lookup loops.

### DIFF
--- a/test-network-function/accesscontrol/suite.go
+++ b/test-network-function/accesscontrol/suite.go
@@ -86,9 +86,11 @@ var _ = ginkgo.Describe(common.AccessControlTestKey, func() {
 				renderedTestCase, err := testConfigure.RenderTestCaseSpec(testcases.Cnf, testType)
 				gomega.Expect(err).To(gomega.BeNil())
 				gomega.Expect(renderedTestCase).ToNot(gomega.BeNil())
-				for _, testCase := range renderedTestCase.TestCase {
-					if !testCase.SkipTest {
-						runTestOnPods(env, testCase, testType)
+
+				// Loop through test cases
+				for i := range renderedTestCase.TestCase {
+					if !renderedTestCase.TestCase[i].SkipTest {
+						runTestOnPods(env, &renderedTestCase.TestCase[i], testType)
 					}
 				}
 			}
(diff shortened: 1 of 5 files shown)


test-network-function/accesscontrol/suite.go
```diff
@@ -189,30 +187,32 @@ func runTestOnPods(env *config.TestEnvironment, testCmd testcases.BaseTestCase,
...
 		lineFields := strings.Split(line, ",")
-		if len(lineFields) != expectedNumFields {
+		//nolint:gomnd
+		if len(lineFields) != 2 {
 			return crNamespaces, fmt.Errorf("failed to parse output line %s", line)
 		}
-		crNamespaces[lineFields[crNameFieldIdx]] = lineFields[namespaceFieldIdx]
+		// 0 = CR Name
+		// 1 = CR Namespace
+		crNamespaces[lineFields[0]] = lineFields[1]
```
greyerof, line 215: I'm a bit confused with this re-hardcoding of slice indexes/lenghts: aren't they magic numbers that should be avoided (as the linter suggests)?

sebrandon1: Yes, I agree. I brought the constants back.

test-network-function/accesscontrol/suite_test.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, tc := range testCases {
+		crMap, err := parseCrOutput(tc.rawOutput)
+		assert.Equal(t, tc.expectedErr, err)
+		if err == nil {
+			assert.Equal(t, tc.expectedOutput, crMap)
+		}
+	}
+}
+
+func TestGetCrsNamespaces(t *testing.T) {
```
edcdavid, line 72: Is it worth keeping this test?

edcdavid (approved):

test-network-function/accesscontrol/suite_test.go
sebrandon1: Why remove it?

greyerof (approved):
I'm not sure why it is necessary to remove helper variables (podName, podNamespace), as they are harmful and they usually allow to make a lot of statements (like printfs) much shorter.

test-network-function/accesscontrol/suite_test.go
edcdavid: no need to remove it, just a suggestion. Already approved

sebrandon1:
I can agree with that.  On the other hand the code doesn't need to allocate new memory (although small and temporary) for doing prints.  If anything, the loop variable `podUnderTest` is too long, and could be shorted to `p`?  Just ideas, maybe for another PR.

test-network-function/accesscontrol/suite_test.go
sebrandon1: Sounds good!

PR #603, opened 2022-01-26
Add IPv6 support for networking tests

edcdavid:
This PR is adding IPV6 support to the ICMP networking tests. 
- The default IP address is now retrieved from the pod manifest instead of using the IP command in the pod 
- Temporarily pointing to the private git@github.com:edcdavid/cnf-certification-test-partner.git ipv6 branch for testing github actions. will be removed before merging
- bugfix: the networking tests are skipped if no network is available for test (e.g. at least one netork with 2 pods and 2 interfaces), instead of skipping when any network is found with less than 2 interfaces to perform the test


pkg/config/config.go
```diff
@@ -84,21 +80,6 @@ func (n NodeConfig) HasDebugPod() bool {
 // DefaultTimeout for creating new interactive sessions (oc, ssh, tty)
 var DefaultTimeout = time.Duration(defaultTimeoutSeconds) * time.Second
 
-// Extract a container IP address for a particular device.  This is needed since container default network IP address
```
jun-rh, line 87: this is no longer needed because we use whereabouts in our local-test-infra now?

edcdavid: I am not sure why getting the IP from the container directly was needed as it should be reported through kubernetes API's status.podIPs (see below since PodIP v1 core). 
I will double check on a openshift 1.6 cluster, maybe it was an older api?


podIPs PodIP arraypatch strategy: mergepatch merge key: ip | podIPs  holds the IP addresses allocated to the pod. If this field is  specified, the 0th entry must match the podIP field. Pods may be  allocated at most 1 value for each of IPv4 and IPv6. This list is empty  if no IPs have been allocated yet.
-- | --

jun-rh: Ok, it's fine if you think it's not needed in general. Just making sure we are not coding the test specific to our local-test-infra.

edcdavid: verified that the object is present in ocp 4.6

test-network-function/networking/suite.go
```diff
@@ -188,23 +200,67 @@ func processContainerIpsPerNet(containerID *configsections.ContainerIdentifier,
...
+	if ip.To4() != nil {
+		return false
+	}
+	return true
+}
+
+func FilterIPListPerVersion(ipList []string, aIPVersion ipVersion) []string {
+	var filteredIPList []string
+	for _, aIP := range ipList {
+		if aIPVersion == IPv4 && IsIPv4(aIP) {
```
greyerof, line 228: No need to change this, but here's a personal opinion on this function: It has too many checks, I think. You could do the same with just one helper function `getIpVersion(ip string) ipVersion` :
```
for _, aIP := range ipList {
    if aIPVersion == getIpVersion(aIP) {
         filteredIpList = append(...)
    }
}


```
The caller should pass a valid IPs list (ipv4/ipv6). In case you really want to throw an error log like in line 233 for invalid ip strings, that can be done at the time where that ipList is built, not here.

pkg/config/autodiscover/pod_info_test.go
```diff
@@ -59,3 +60,54 @@ func TestPodGetAnnotationValue(t *testing.T) {
...
+
+	// Passing tests
+	var testsExpectFail = []test{
+		{name: "ipv4ipv6",
+			testFile: "testorchestrator.json",
+			want:     []string{"2.2.2.3", "fd00:10:244:1::3"},
+		},
+	}
+
+	// Failing tests
```
greyerof, line 79: Failing? I found this a bit misleading, since all unit tests must pass.

pkg/config/autodiscover/pod_info_test.go
```diff
@@ -59,3 +60,54 @@ func TestPodGetAnnotationValue(t *testing.T) {
...
+			testFile: "ipv6pod.json",
+			want:     []string{"fd00:10:244:1::3"},
+		},
+	}
+
+	// Expect fail
+	for _, tt := range testsExpectFail {
+		t.Run(tt.name, func(t *testing.T) {
+			pr := loadPodResource(path.Join(filePath, tt.testFile))
+			if got := pr.getDefaultPodIPAddresses(); reflect.DeepEqual(got, tt.want) {
```
greyerof, line 99: I understand the other cases, but what are you actually testing here?

test-network-function/networking/suite.go
edcdavid: fixed

hamadise (approved):

PR #604, opened 2022-01-27
unitTest for file suite.go in "networking"

shimritproj:
unit tests for file suite.go in "networking".
add test cases that check cases for test that check if all the listening port is declared.


test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestParseVariables(t *testing.T) {
+	// expected inputs
+	testCases := []struct {
+		// inputs
+		inputRes      string
+		declaredPorts map[key]string
+
+		// expected outputs here
+		expectedDeclaredPorts map[key]string
+		exceptedRes           string
```
sebrandon1, line 35:
```suggestion
		expectedRes           string
```

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}{
+		{
+			inputRes:              `[\n  {\n    \"containerPort\": 8080,\n    \"name\": \"http-probe\",\n    \"protocol\": \"TCP\"\n  }\n]`,
+			declaredPorts:         map[key]string{},
+			expectedDeclaredPorts: map[key]string{},
+			exceptedRes:           `[\n  {\n    \"containerPort\": 8080,\n    \"name\": \"http-probe\",\n    \"protocol\": \"TCP\"\n  }\n]`,
+		},
+	}
```
sebrandon1, line 43: We should have some more testcases here like what happens if the `inputRes` is empty string.

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		// expected outputs here
+		expectedlisteningPorts map[key]string
+	}{
+		{
+			commandlisten:          []string{"nsenter -t 4380 -n", "ss -tulwnH"},
+			nodeOc:                 nil,
+			listeningPorts:         map[key]string{},
+			expectedlisteningPorts: map[key]string{},
+		},
```
sebrandon1, line 96: Test case with populated `listeningPorts`?

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+		declaredPorts  map[key]string
+
+		// expected outputs here
+		expectedres map[key]string
+	}{
+		{
+			listeningPorts: map[key]string{},
+			declaredPorts:  map[key]string{},
+			expectedres:    map[key]string{},
+		},
```
sebrandon1, line 119: Test cases with populated values?

sebrandon1 (changes requested):
Just some comments about adding more test cases to the array of tests.

sebrandon1:
Thanks!

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+			container:             0,
+			podName:               "",
+			podNamespace:          "",
+			declaredPorts:         map[key]string{},
+			expectedDeclaredPorts: map[key]string{},
+		},
+	}
+
+	for _, tc := range testCases {
+		err := declaredPortList(tc.container, tc.podName, tc.podNamespace, tc.declaredPorts)
```
greyerof, line 93: This UT won't work unless you mock, for each test case, the utils.ExecuteCommand() function that declarePortList() is calling internally. Other than that, I don't see the point of having this UT, apart from testing a possible error thrown by the utils.ExecuteCommand() when parsing the output.

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func TestDeclaredPortList(t *testing.T) {
+	// expected inputs
+	testCases := []struct {
+		// inputs
+		container     int
+		podName       string
+		podNamespace  string
+		declaredPorts map[key]string
```
greyerof, line 71: See my other comment about this declaredPorts field.

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+		{
+			commandlisten:          []string{},
+			nodeOc:                 nil,
+			listeningPorts:         map[key]string{},
+			expectedlisteningPorts: map[key]string{},
+		},
+	}
+	for _, tc := range testCases {
+		err := listeningPortList(tc.commandlisten, tc.nodeOc, tc.listeningPorts)
```
greyerof, line 124: Pretty much the same as the previous UT.

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+			expectedres:    map[key]string{},
+		},
+		{
+			listeningPorts: map[key]string{{port: 8080, protocol: "TCP"}: ""},
+			declaredPorts:  map[key]string{},
+			expectedres:    map[key]string{{port: 8080, protocol: "TCP"}: ""},
+		},
+	}
+	for _, tc := range testCases {
+		err := checkIfListenIsDeclared(tc.listeningPorts, tc.declaredPorts)
```
greyerof, line 157: The checkIfListenIsDeclared() function does never return an error. It's always returning an empty map, which will never be nil. That also makes the assertion on line 158 useless, since it will never fail. I suggest changing this to:
```
listeningPortsNotDeclared := checkIfListenIsDeclared(tc.listeningPorts, tc.declaredPorts)
```

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+		assert.Nil(t, err)
```
greyerof, line 158: As said in previous comment, this assertion is wrong. You should check that "len(listeningPortsNotDeclared) == tc.expectedBadPorts" or similar, based on each test case.

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+		assert.Equal(t, tc.listeningPorts, checkIfListenIsDeclared(nil, nil))
```
greyerof, line 159: Whay are you calling checkIfListenIsDeclared(nil, nil) here? This assertion will only pass in case tc.listeningPorts is an empty map for each test case. Is that what you really want here?

greyerof (commented):
A couple of things about this PR:
	1. All of these UTs are currently failing when executed locally. You can use vs code to debug it, if needed.
	2. The `grep -v "github.com/test-network-function/test-network-function/test-network-function" `in the Makefile is preventing this UTs to be run. I think we need to update this to allow unit tests on the suites to be executed.

test-network-function/networking/suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/test-network-function/test-network-function/pkg/tnf/interactive"
+)
+
+func TestParseVariables(t *testing.T) {
+	// expected inputs
+	testCases := []struct {
+		// inputs
+		inputRes      string
```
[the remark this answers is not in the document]
shimritproj: I was add note that explain what this.

test-network-function/networking/suite_test.go
```diff
@@ -120,7 +125,15 @@ func TestListeningPortList(t *testing.T) {
 			expectedlisteningPorts: map[key]string{},
 		},
 	}
+	origFunc := utils.ExecuteCommand
+	defer func() {
+		utils.ExecuteCommand = origFunc
+	}()
 	for _, tc := range testCases {
+		utils.ExecuteCommand = func(command string, timeout time.Duration, context *interactive.Context) (string, error) {
+			output, err := os.ReadFile(tc.jsonFileName)
```
greyerof, line 134: This UT is still failing. You have to initialize the tc.jsonFileName for each TC (or make it a global variable if all TCs can use the same file) on this unit test as you did on TestDeclaredPortList.

pkg/config/autodiscover/testdata/pods_with_json.json
```diff
@@ -0,0 +1,1905 @@
+{
```
greyerof, line 1: This file does not belong to the "autodiscover" package. You should create a "testdata" folder inside the **networking** test suite folder and put it there, because these UTs are testing functions of the networking package. Also, it shouldn't have the full pod/s object: only the ports array, because that's the content of the "res" variable that the declarePortList function will use to call the parseVariables one:
```
[
    {
        "containerPort": 8080,
        "name": "http-probe",
        "protocol": "TCP"
    }
]
```
The file name should be something like "test_ports.json" instead.

test-network-function/networking/suite_test.go
```diff
@@ -17,40 +17,40 @@
...
 		},
 		{
-			inputRes:              `[\n  {\n    \"containerPort\": 8080,\n    \"name\": \"http-probe\",\n    \"protocol\": \"TCP\"\n  }\n]`,
+			inputRes:              "[\n  {\n    \"containerPort\": 8080,\n    \"name\": \"http-probe\",\n    \"protocol\": \"UDP\"\n  }\n]",
 			declaredPorts:         map[key]string{},
-			expectedDeclaredPorts: map[key]string{{port: 8080, protocol: "TCP"}: "http-probe"},
-			expectedRes:           `[\n  {\n    \"containerPort\": 8080,\n    \"name\": \"http-probe\",\n    \"protocol\": \"TCP\"\n  }\n]`,
+			expectedDeclaredPorts: map[key]string{{port: 8080, protocol: "UDP"}: "http-probe"},
+			expectedRes:           "[\n  {\n    \"containerPort\": 8080,\n    \"name\": \"http-probe\",\n    \"protocol\": \"UDP\"\n  }\n]",
 		},
```
greyerof, line 54: What about adding a new TC here with more than one expected ports?

shimritproj: yes, I was add.

test-network-function/networking/suite_test.go
shimritproj: is works now.

test-network-function/networking/suite_test.go
shimritproj: I fixed it.

test-network-function/networking/suite_test.go
shimritproj: I fixed it and this works now.

test-network-function/networking/suite_test.go
shimritproj: you're right! I fixed it.

pkg/config/autodiscover/testdata/pods_with_json.json
shimritproj: I was add 2 files: one for test for declarePort and one for test listeningPort.

shimritproj: The name of file for "declarePort" is "test_ports.json" and the name of file for "listeningPort" is "test_listening_port.json". is Ok? @greyerof

test-network-function/networking/suite_test.go
shimritproj: In this function ,I replaced to :
assert.Equal(t, len(res), 0)
assert.Equal(t, len(tc.listeningPorts), len(tc.declaredPorts))

and this works now.
@greyerof

test-network-function/networking/suite_test.go
```diff
@@ -17,10 +17,176 @@
...
+	defer func() {
+		utils.ExecuteCommand = origFunc
+	}()
+	for _, tc := range testCases {
+		utils.ExecuteCommand = func(command string, timeout time.Duration, context *interactive.Context) (string, error) {
+			output, err := os.ReadFile(tc.jsonFileName)
+			return string(output), err
+		}
+		err := listeningPortList(tc.commandlisten, tc.nodeOc, tc.listeningPorts)
+		assert.Nil(t, err)
```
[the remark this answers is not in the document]
shimritproj: This is incorrect because the input listeningPorts is not the same as the output(expectedListeningPorts).

test-network-function/networking/testdata/test_ports.json
```diff
@@ -0,0 +1,22 @@
...
+},
+{
+    "containerPort": 8080,
+    "name": "http-probe",
+    "protocol": "TCP"
+},
+{
+    "containerPort": 8080,
+    "name": "http-probe",
+    "protocol": "TCP"
```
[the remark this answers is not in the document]
shimritproj: I know, I did it on purpose to see that he knows how to deal with duplicates.

test-network-function/networking/suite_test.go
```diff
@@ -17,10 +17,176 @@
...
+
+func TestCheckIfListenIsDeclared(t *testing.T) {
+	// expected inputs
+	testCases := []struct {
+		// inputs
+		listeningPorts map[key]string
+		declaredPorts  map[key]string
+
+		// expected outputs here
+		expectedres map[key]string
```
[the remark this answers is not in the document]
shimritproj: It's not true because he keeps all the ports they listening but has not declared port.

test-network-function/networking/suite_test.go
```diff
@@ -17,10 +17,176 @@
...
+		{
+			listeningPorts: map[key]string{{port: 8080, protocol: "TCP"}: ""},
+			declaredPorts:  map[key]string{{port: 8080, protocol: "TCP"}: "http-probe"},
+			expectedres:    map[key]string{},
+		},
+	}
+	for _, tc := range testCases {
+		res := checkIfListenIsDeclared(tc.listeningPorts, tc.declaredPorts)
+		assert.Equal(t, len(res), 0)
+		assert.Equal(t, len(tc.listeningPorts), len(tc.declaredPorts))
```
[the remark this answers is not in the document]
shimritproj: I think that this is good test and about your comments:
a+b) its not correct becuase have situation that res is not same to listening port and this is OK. res incluse all the port that listening **but** no declared.
c)I did it in this TC: 
{
			listeningPorts: map[key]string{{port: 8080, protocol: "TCP"}: ""},
			declaredPorts:  map[key]string{{port: 8080, protocol: "TCP"}: "http-probe"},
			expectedres:    map[key]string{},
		},
e)I did it in this TC:
{
			listeningPorts: map[key]string{},
			declaredPorts:  map[key]string{},
			expectedres:    map[key]string{},
		},

pkg/config/autodiscover/testdata/pods_with_json.json
greyerof: Yes, that name is ok for me.

test-network-function/networking/testdata/test_ports.json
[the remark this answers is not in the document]
greyerof: Ok

test-network-function/networking/suite_test.go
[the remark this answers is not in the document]
shimritproj: I fixed it.

PR #616, opened 2022-02-03
More typos

sebrandon1:
Found some more typos.

### DIFF
--- a/pkg/tnf/handlers/podsets/podsets.go
+++ b/pkg/tnf/handlers/podsets/podsets.go
@@ -35,9 +35,9 @@ type PodSet struct {
 	Replicas    int
 	Ready       int
 	UpToDate    int
-	Available   int // this param exist in deployment and not exsit on statefulset
+	Available   int // this param applies only to deployments, not to statefulsets
 	Unavailable int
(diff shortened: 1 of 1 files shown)


pkg/tnf/handlers/podsets/podsets.go
```diff
@@ -35,9 +35,9 @@ type PodSet struct {
 	Replicas    int
 	Ready       int
 	UpToDate    int
-	Available   int // this param exist in deployment and not exsit on statefulset
+	Available   int // this param exist in deployment and not exist on statefulset
```
greyerof, line 38: Shouldn't it be changed to "this param **exists** in deployment _but_ **does not exist** on statefulset" to be more correct in english? Anyway, I prefer the "apply" verb here: "this param **applies** only to deployments, not to statefulsets", because the param will always exist in this struct, but the value will be always 0 for statefulsets, right?

sebrandon1: Agreed.  I will adjust the strings.

greyerof (approved):

edcdavid (approved):

PR #620, opened 2022-02-04
Pushing latest release only once

edcdavid:
Push latest release only once at the end of the process of building each release in the matrix, instead of after each individual release.


.github/workflows/tnf-image.yaml
```diff
@@ -157,4 +158,9 @@ jobs:
           password: ${{ secrets.QUAY_ROBOT_TOKEN }}
 
       - name: Push the newly built image to Quay.io
-        run: docker push --all-tags ${REGISTRY}/${IMAGE_NAME}
+        run: |
+          docker push ${REGISTRY}/${IMAGE_NAME}:${TNF_VERSION}
+          if [ "$CURRENT_VERSION_GENERIC_BRANCH" == "$LATEST_BRANCH_VERSION" ]; then
+            docker push ${REGISTRY}/${IMAGE_NAME}:${IMAGE_TAG}
```
greyerof, line 164: This will speed up this workflow, but it seems a bit opposing to the original idea of it, which is to update images in case the base UBI image has security patches/updates. 
```
  ...
  # This is useful for keeping the image up-to-date with security
  # patches provided in the UBI.
  ...
```
With this PR, the UBI images of 3.1.x, 3.2.x... won't be updated. Do we really want this?

greyerof: Ok, thanks for clarifying.

sebrandon1 (approved):

Issue #621, opened 2022-02-07
IPv6 partner repo update broke image generation


edcdavid:
Image generation is broken because the name of the action used to start the cluster has been renamed to "start-k8s-cluster" from "start-minikube-cluster" .
We need to add some logic in the cnf test action to use the old name before version 3.3.0

PR #622, opened 2022-02-08
Fix image generation after ipv6 merge

edcdavid:
Uses available github action to build the test cluster so it should work for partner repo before and after 3.3.0
Fix for: https://github.com/test-network-function/test-network-function/issues/621

### DIFF
--- a/.github/workflows/tnf-image.yaml
+++ b/.github/workflows/tnf-image.yaml
@@ -126,10 +126,33 @@ jobs:
           repository: test-network-function/cnf-certification-test-partner
           path: cnf-certification-test-partner
           ref: ${{ env.PARTNER_VERSION }}
-          
-      - name: Start the minikube cluster for `local-test-infra`
+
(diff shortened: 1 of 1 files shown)


.github/workflows/tnf-image.yaml
```diff
@@ -126,10 +126,33 @@ jobs:
           repository: test-network-function/cnf-certification-test-partner
           path: cnf-certification-test-partner
           ref: ${{ env.PARTNER_VERSION }}
-          
-      - name: Start the minikube cluster for `local-test-infra`
+
+      # Only one of the following 2 steps will work depending on the repo version
+
+      # For versions > 3.3.0
```
hamadise, line 132: this part here should work for older code as well ? did you tried it and it did not work ?

edcdavid: yes. It works because it is checking the presence of the file first

hamadise (approved):

sebrandon1 (approved):

PR #638, opened 2022-02-21
ModuleInTree function modified to use /sys/module/<mod>/taint

greyerof:
The use of modinfo won't work in cases where modules have been loaded
from inside pods (like in SRO), since those pods have a different
filesystem namespace for /lib/modules.

The ModuleInTree function has been updated to parse the content of the
file /sys/module/<module>/taint, which returns a string with a letter
per taint. The taint <-> letter correspondence list can be found here:
- https://github.com/torvalds/linux/blob/master/kernel/panic.c#L369
- https://www.kernel.org/doc/html/latest/admin-guide/tainted-kernels.html

Some log traces improved.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -187,17 +187,16 @@ func GetModulesFromNode(nodeName string, nodeOc *interactive.Oc) []string {
 	return strings.Split(strings.ReplaceAll(output, "\r\n", "\n"), "\n")
 }
 
+// ModuleInTree returns true if the module hasn't tainted the kernel with the
+// out-of-tree ("O") bit. The /sys/module/<module>/taint file has the stringified
+// values of all the taints it is adding to the kernel. Each bit is one letter.
+// Refs:
+//  1. https://www.kernel.org/doc/html/latest/admin-guide/tainted-kernels.html
+//  2. https://github.com/torvalds/linux/blob/master/kernel/panic.c#L369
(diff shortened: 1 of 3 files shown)


sebrandon1 (approved):

sebrandon1:
I will also update my PR [here](https://github.com/test-network-function/cnf-certification-test/pull/16) to use this new path to check for modules.

hamadise (approved):

pkg/utils/utils.go
```diff
@@ -187,17 +187,16 @@ func GetModulesFromNode(nodeName string, nodeOc *interactive.Oc) []string {
...
 	cmdOutput := RunCommandInNode(nodeName, nodeOc, command, timeoutPid)
-	outputSlice := strings.Split(strings.ReplaceAll(cmdOutput, "\r\n", "\n"), "\n")
-	// The output, if found, should look something like 'intree:   Y'.
-	// As long as we look for 'intree:' being contained in the string we should be good to go.
-	found := false
-	if StringInSlice(outputSlice, `intree:`, false) {
-		found = true
-	}
-	return found
+	return !strings.Contains(cmdOutput, "O")
```
edcdavid, line 199: Is the Taint OOT superseding the other 2 module taints below? If not we could add decoding fro these
[ TAINT_PROPRIETARY_MODULE ]	= { 'P', 'G', true },
[ TAINT_OOT_MODULE ]		= { 'O', ' ', true },
[ TAINT_UNSIGNED_MODULE ]	= { 'E', ' ', true },

PR #644, opened 2022-03-07
Added timeout const for node drain steps, increased drainTimeoutMinutes

shaior:
We used scalingTimeout var in the pod-recreation test, I believe it should be a different var, 
so I added nodeDrainTimeout. 
Also increased drainTimeoutMinutes to 7.


hamadise (approved):

hamadise:
I think what you can do, is a multiplier of scalingtimeout, this way a single env is needed.

greyerof (approved):

PR #649, opened 2022-03-10
Get all CSV's installPlans

greyerof:
As CSV/Operators can be installed through several installPlans, this
PR iterates through all of them saving the needed fields.


sebrandon1 (approved):

ramperher:
I think my approval is not enough to have the 2 approvals, because I don't have write permissions, so one more person from your team would need to do a review to merge this, @greyerof

edcdavid (approved):

PR #651, opened 2022-03-11
Removing extra statefulset test

edcdavid:
This is the second part of the partner PR to improve testing stateful sets. Deployment and statefulset are now created (and exist) at the same time 
Before, we were testing only deployments during the smoke test run not using image and 
we were only using the statefulset (no deployment) during the smoke test run with the image 
This PR remove the code deleting the deployment and creating the statefulset to test the image since both statefulset and deployment are always created at the same time.
Will remove reference to pod-recreation repo before mergingno deployments
https://github.com/test-network-function/cnf-certification-test-partner/pull/90


hamadise (approved):

greyerof:
David, please, can you provide more info about why is this needed?

greyerof:
David, please, can you provide more info about why is this needed?

sebrandon1 (approved):

PR #653, opened 2022-03-14
Bug fix: kernel tainted test did not fail on other types of failures

sebrandon1:
As I was [writing unit tests](https://github.com/test-network-function/cnf-certification-test/pull/63) for the "new" repo, I came across this bug where we would only flag the node as failed if it failed because of a `module was loaded` reason.

### DIFF
--- a/test-network-function/platform/suite.go
+++ b/test-network-function/platform/suite.go
@@ -415,18 +415,20 @@ func testTainted(env *config.TestEnvironment) {
 				}
 				taintMsg, individualTaints := decodeKernelTaints(taintedBitmap)
 
-				// We only will fail the tainted kernel check if the reason for the taint
-				// only pertains to `module was loaded`.
+				// Count how many taints come from `module was loaded` taints versus `other`
 				log.Debug("Checking for 'module was loaded' taints")
-				moduleCheck := false
+				log.Debug("individualTaints", individualTaints)
+				moduleCheckCtr := 0
+				otherTaintCtr := 0
 				for _, it := range individualTaints {
 					if strings.Contains(it, `module was loaded`) {
-						moduleCheck = true
-						break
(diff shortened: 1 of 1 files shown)


hamadise (approved):

test-network-function/platform/suite.go
```diff
@@ -441,6 +441,8 @@ func testTainted(env *config.TestEnvironment) {
 					// Looks through the accepted taints listed in the tnf-config file.
 					// If all of the tainted modules show up in the configuration file, don't fail the test.
 					nodeTaintsAccepted = taintsAccepted(env.Config.AcceptedKernelTaints, taintedModules)
+				} else if len(individualTaints) > 0 {
```
greyerof, line 444: What if we have "module was loaded" plus other not-related to out-of-tree module taints? The white list mechanism would allow those taints to pass this TC.

sebrandon1: True.  I'll play around with it more.

sebrandon1: Okay, I added counters to always mark the node as `failed` if there are other tainted reasons compared to `module was loaded`.

test-network-function/platform/suite.go
```diff
@@ -443,6 +445,11 @@ func testTainted(env *config.TestEnvironment) {
 					nodeTaintsAccepted = taintsAccepted(env.Config.AcceptedKernelTaints, taintedModules)
 				}
 
+				// If there are other taints than module was loaded, set the result to false/fail.
+				if otherTaintCtr > 0 {
```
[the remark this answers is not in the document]
sebrandon1: Yeah that works.  Or else I can use my taint counters and print them to the log.  I'll play with it.

greyerof (approved):

PR #657, opened 2022-03-22
Adding data about number of container

shimritproj:
Adding data about number of container.


sebrandon1 (commented):
Could we also put the container's name in addition to the PID?

PR #658, opened 2022-03-22
Adding data about number of container

shimritproj:


sebrandon1:
After looking at this today, @shimritproj and I found out that the `nsenter` command that accepts a PID doesn't get LISTEN ports that are only attached to that PID.  It shows all of the same ports that the host has.

```
$ ss -tulwnHp | grep LISTEN
tcp LISTEN 0      0             0.0.0.0:5355  0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:111   0.0.0.0:*
tcp LISTEN 0      0          127.0.0.53:53    0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:22    0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:57239 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:34171 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:2049  0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:59681 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:59461 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:39207 0.0.0.0:*
tcp LISTEN 0      0           127.0.0.1:10248 0.0.0.0:*
tcp LISTEN 0      0           127.0.0.1:36841 0.0.0.0:*
tcp LISTEN 0      0                   *:5355        *:*
tcp LISTEN 0      0                   *:111         *:*
tcp LISTEN 0      0                   *:10256       *:*
tcp LISTEN 0      0                   *:22          *:*
tcp LISTEN 0      0                   *:42807       *:*
tcp LISTEN 0      0                   *:2049        *:*
tcp LISTEN 0      0                   *:52675       *:*
tcp LISTEN 0      0                   *:32837       *:*
tcp LISTEN 0      0                   *:39079       *:*
tcp LISTEN 0      0                   *:60935       *:*
tcp LISTEN 0      0                   *:2376        *:*
tcp LISTEN 0      0                   *:10249       *:*
tcp LISTEN 0      0                   *:10250       *:*
```

```
$ sudo nsenter -t 69438 -n ss -tulwnH | grep LISTEN
tcp LISTEN 0      0             0.0.0.0:5355  0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:111   0.0.0.0:*
tcp LISTEN 0      0          127.0.0.53:53    0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:22    0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:57239 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:34171 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:2049  0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:59681 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:59461 0.0.0.0:*
tcp LISTEN 0      0             0.0.0.0:39207 0.0.0.0:*
tcp LISTEN 0      0           127.0.0.1:10248 0.0.0.0:*
tcp LISTEN 0      0           127.0.0.1:36841 0.0.0.0:*
tcp LISTEN 0      0                   *:5355        *:*
tcp LISTEN 0      0                   *:111         *:*
tcp LISTEN 0      0                   *:10256       *:*
tcp LISTEN 0      0                   *:22          *:*
tcp LISTEN 0      0                   *:42807       *:*
tcp LISTEN 0      0                   *:2049        *:*
tcp LISTEN 0      0                   *:52675       *:*
tcp LISTEN 0      0                   *:32837       *:*
tcp LISTEN 0      0                   *:39079       *:*
tcp LISTEN 0      0                   *:60935       *:*
tcp LISTEN 0      0                   *:2376        *:*
tcp LISTEN 0      0                   *:10249       *:*
tcp LISTEN 0      0                   *:10250       *:*
```

greyerof (commented):
The whole implementation of this TC in this repo is wrong, so I believe just changing this log trace won't help much. The original intention of this TC is checking that each container of the pod is not listening to any port that hasn't been declared in its own container spec. But, the current code here has the following important problems:
- Iterates through pods under test, but it accumulates the declared ports that it finds for each container of the pod inside the same map.
- Gets the listening ports for the first container of the pod only (line 498) and compares those listening ports against the map for the ports of all the containers in the pod.
- It never clears both maps used to accumulate declared/listening ports, so it's mixing the ports from all the containers of all the pods under test in the cluster.
All this issues have been addressed and fixed in the new repo, so we should implement those fixes here as well.

greyerof:
Yet, somehow, the nsenter command works here, @sebrandon1 . Look at the tnf output:
```
DEBUG  [Mar 23 09:17:42.119][utils.go: 142] Command handler's COMMAND string value: nsenter -t 1480315 -n  ss -tulwnH 
DEBUG  [Mar 23 09:17:42.127][reel.go: 216] command status: output=tcp LISTEN 0      4096   *:8080 *:*, match=tcp LISTEN 0      4096   *:8080 *:*, outputStatus=0, matchStatus=0, caseIndex=0 
```

That `tcp LISTEN 0      4096   *:8080 *:*` is the liveness/readiness probe port that [the container's app](https://github.com/test-network-function/cnf-certification-test-partner/blob/main/testapp/cmd/web/main.go) is listening.

shimritproj:
I think the problem also exists in the new repo because also in the new repo I use the command "ss -tulwnH"

And as Brandon wrote here, it always returns the same output.

sebrandon1 (dismissed):
I think this looks good.  Should help by moving those variable declarations into the loop.

test-network-function/networking/suite.go
```diff
@@ -450,73 +450,66 @@ func listeningPortList(commandlisten []string, nodeOc *interactive.Context, list
...
-					continue OUTER
-				}
+		for _, cut := range env.ContainersUnderTest {
+			declaredPorts := make(map[key]bool)
+			listeningPorts := make(map[key]bool)
+			container, _ := strconv.Atoi(cut.ContainerUID)
+			err := declaredPortList(container, cut.PodName, cut.Namespace, declaredPorts)
+			if err != nil {
+				tnf.ClaimFilePrintf("Failed to get declared port for container %d due to %v, skipping pod %s", cut.ContainerUID, err, cut.PodName)
+				continue
```
greyerof, line 483: 1. Change the ClaimFilePrinft message to:
```
tnf.ClaimFilePrintf("Failed to get declared ports list for container %s pod %s ns %s, err %s", cut.ContainerName, cut.PodName, cut.Namespace, err)
```
2. Add this container to the failedContainers slice before "continue".

test-network-function/networking/suite.go
```diff
@@ -450,73 +450,66 @@ func listeningPortList(commandlisten []string, nodeOc *interactive.Context, list
...
-			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, x.ContainerUID, x.ContainerRuntime)
+			containerPID := utils.GetContainerPID(nodeName.GetNodeName(), nodeOc, cut.ContainerUID, cut.ContainerRuntime)
 
 			commandlisten := []string{utils.AddNsenterPrefix(containerPID), commandportlisten}
 
 			err = listeningPortList(commandlisten, nodeOc.Context, listeningPorts)
 			if err != nil {
-				tnf.ClaimFilePrintf("Failed to get listening port for pod name %s in pod namespace %s due to %v, skipping this pod", podUnderTest.Name, podUnderTest.Namespace, err)
+				tnf.ClaimFilePrintf("Failed to get listening port for pod name %s in pod namespace %s on container %s due to %v, skipping this pod", cut.PodName, cut.Namespace, cut.ContainerUID, err)
+				failedcontainers = append(failedcontainers, fmt.Sprintf("podName: %s, Namespace: %s, container:%s", cut.PodName, cut.Namespace, cut.ContainerUID))
```
greyerof, line 498: Please, do not use cut.ContainerUID in these error traces, as it's just a number without a meaning for the customer. Change this trace to:
```
tnf.ClaimFilePrintf("Failed to get listening ports for container %s pod %s ns %s due to err %s", cut.ContainerName, cut.PodName, cut.Namespace, err)
```

test-network-function/networking/suite.go
```diff
@@ -450,73 +450,66 @@ func listeningPortList(commandlisten []string, nodeOc *interactive.Context, list
...
 				continue
 			}
 			// compare between declaredPort,listeningPort
-			undeclaredPorts = checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			undeclaredPorts := checkIfListenIsDeclared(listeningPorts, declaredPorts)
 			for k := range undeclaredPorts {
-				tnf.ClaimFilePrintf("The port %d on protocol %s in pod name %s and pod namespace is %s not declared.", k.port, k.protocol, podUnderTest.Name, podUnderTest.Namespace)
+				tnf.ClaimFilePrintf("The port %d on protocol %s in pod name %s and namespace is %s not declared on container %s.", k.port, k.protocol, cut.PodName, cut.Namespace, containerPID)
```
greyerof, line 504: Use `cut.ContainerName` instead of `containerPID` in this trace, please.

test-network-function/networking/suite.go
```diff
@@ -450,73 +450,66 @@ func listeningPortList(commandlisten []string, nodeOc *interactive.Context, list
...
 			}
 			if len(undeclaredPorts) != 0 {
-				failedPods = append(failedPods, *podUnderTest)
+				failedcontainers = append(failedcontainers, fmt.Sprintf("podName: %s, Namespace: %s, container:%s", cut.PodName, cut.Namespace, cut.ContainerUID))
```
greyerof, line 507: Use `cut.ContainerName` instead of `cut.ContainerUID`, please.

test-network-function/networking/suite.go
```diff
@@ -450,73 +450,66 @@ func listeningPortList(commandlisten []string, nodeOc *interactive.Context, list
...
 	return res
 }
 
 func testListenAndDeclared(env *config.TestEnvironment) {
-	declaredPorts := make(map[key]string)
-	listeningPorts := make(map[key]string)
-	undeclaredPorts := make(map[key]string)
-	var skippedPods []configsections.Pod
-	var failedPods []configsections.Pod
+	var failedcontainers []string
```
greyerof, line 473: Rename variable to `failedContainers`, please.

greyerof:
My previous comment was wrong, sorry so much, @shimritproj . The old implementation was better than the one in the new repo, as all the containers share the network namespace so there's no need to call the listeningPortList() for each container in the pod.

We'll need to change the implementation in the new repo and make just some fixes in this old one. Basically, the problems that the old repo's implementation had were:
- The maps should be declared and initialized as empty an each pod iteration:
```
        ...
	OUTER:
		for _, podUnderTest := range env.PodsUnderTest {
			declaredPorts := make(map[key]bool)
			listeningPorts := make(map[key]bool)
```
- If the listeningPortList() function returns err != nil, then the pod needs to be added to the skipped list before the "continue". Otherwise, the tc would pass.
```
	err = listeningPortList(commandlisten, nodeOc.Context, listeningPorts)
	if err != nil {
		tnf.ClaimFilePrintf("Failed to get listening port for pod name %s in pod namespace %s due to %v, skipping this pod", podUnderTest.Name, podUnderTest.Namespace, err)
		skippedPods = append(skippedPods, *podUnderTest)
		continue
	}
```

test-network-function/networking/suite.go
```diff
@@ -503,10 +501,11 @@ func testListenAndDeclared(env *config.TestEnvironment) {
...
 			if err != nil {
 				tnf.ClaimFilePrintf("Failed to get listening port for pod name %s in pod namespace %s due to %v, skipping this pod", podUnderTest.Name, podUnderTest.Namespace, err)
+				skippedPods = append(skippedPods, *podUnderTest)
 				continue
 			}
 			// compare between declaredPort,listeningPort
-			undeclaredPorts = checkIfListenIsDeclared(listeningPorts, declaredPorts)
+			undeclaredPorts := checkIfListenIsDeclared(listeningPorts, declaredPorts)
 			for k := range undeclaredPorts {
 				tnf.ClaimFilePrintf("The port %d on protocol %s in pod name %s and pod namespace is %s not declared.", k.port, k.protocol, podUnderTest.Name, podUnderTest.Namespace)
```
greyerof, line 510: In order to make this trace easier to understand for the customer when this test case fails, I suggest this change:
```
tnf.ClaimFilePrintf("pod %s ns %s is listening on port %d protocol %d, but that port was not declared in any container spec.",
    podUnderTest.Name, podUnderTest.Namespace, k.port, k.protocol)
)
```
As there's no way (yet) to know which container of the pod actually opened that port, that's the most reasonable message we can print. What do you think, @edcdavid & @sebrandon1 ?

greyerof (commented):
Good job, I just added a minor change in a trace. 
Now we need to change the implementation in the new repo to be similar as this one.
Sorry again for the inconveniences caused by my first reviews on this TC.

test-network-function/networking/suite.go
shimritproj: I agree.

greyerof (approved):

test-network-function/networking/suite.go
edcdavid: I agree too

sebrandon1 (approved):

PR #666, opened 2022-04-12
Diagnostic mode fix: close debug oc sessions before removing node lab…

greyerof:
…els.

This commit tries to fix the crash that happens when running the
container in diagnostic mode: race condition when the program finishes,
as the oc debug session were not closed before the labels are being
removed from the nodes.


edcdavid (approved):
We should revert the timeout value to previous 20 seconds to make sure this solves the issue

hamadise (approved):

sebrandon1 (approved):

